PR #101, opened 2020-08-30
Added UI tests for Showkase Browser

### LINKED ISSUE #41: Add UI Tests for the browser

vinaygaba:
Resolves #41

@airbnb/showkase-maintainers

### DIFF
--- a/showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTest.kt
+++ b/showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTest.kt
@@ -47,7 +47,7 @@ class ShowcaseBrowserTest {
         clickRowWithText("Components")
 
         // Verify that all the groups are displayed on the screen
-        verifyRowsWithTextAreDisplayed(listOf("Group1", "Group2", "Group3"))
+        verifyRowsWithTextAreDisplayed("Group1", "Group2", "Group3")
     }
 
     @Test
@@ -59,7 +59,7 @@ class ShowcaseBrowserTest {
         clickRowWithText("Colors")
 
         // Verify that a row for the group "Light Colors" is visible on the screen
-        verifyRowsWithTextAreDisplayed(listOf("Light Colors"))
+        verifyRowsWithTextAreDisplayed("Light Colors")
     }
 
     @Test
@@ -71,7 +71,7 @@ class ShowcaseBrowserTest {
         clickRowWithText("Typography")
 
         // Verify that all typography groups are visible on the screen
-        verifyRowsWithTextAreDisplayed(listOf("Material", "Holo"))
+        verifyRowsWithTextAreDisplayed("Material", "Holo")
     }
 
     @Test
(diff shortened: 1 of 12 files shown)


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -98,7 +98,8 @@ class ShowkaseProcessor: AbstractProcessor() {
     private fun processComponentAnnotation(roundEnvironment: RoundEnvironment)
             : Set<ShowkaseMetadata> {
         val showkaseComposablesMetadata = processShowkaseAnnotation(roundEnvironment)
-        val previewComposablesMetadata = processPreviewAnnotation(roundEnvironment)
+//        val previewComposablesMetadata = processPreviewAnnotation(roundEnvironment)
```
vinaygaba, line 101: I did this just to be able to run the tests. The `processPreviewAnnotation` method crashes when running a test. This is because the `Class.forName(PREVIEW_CLASS_NAME)` throws a `ClassNotFoundException` :( I tried added the `RuntimeJar` to the classpath of androidTestImplementation(just like how I do it for testImplementation) but that doesn't solve the problem unfortunately!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -117,19 +117,18 @@ class ShowkaseProcessor: AbstractProcessor() {
...
-            .asType()
-        return roundEnvironment.getElementsAnnotatedWith(previewClassAnnotation).mapNotNull { element ->
-            showkaseValidator.validateComponentElement(element, composableTypeMirror, typeUtils, 
-                previewClass.simpleName)
-            val showkaseMetadata = getShowkaseMetadataFromPreview(
-                element as ExecutableElement, elementUtils, typeUtils, previewTypeMirror, showkaseValidator
-            )
-            showkaseMetadata
-        }.toSet()
+        val previewTypeElement = elementUtils.getTypeElement(PREVIEW_CLASS_NAME)
```
vinaygaba, line 120: This was updated because using `Class.forName` directly was causing a `ClassNotFoundException`. After asking around, I was advised to directly use `elementUtils.getTypeElement` instead of using the canonical name of `Class.forName` which effectively does the same thing. 

However, I needed to add a null check to this previewTypeElement. All this is working fine even with the null check but I'm wondering in which situations does it return null and why? If I remove the null check, I get an NPE.

showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTestFlows.kt
```diff
@@ -0,0 +1,77 @@
...
+
+internal fun clickRowWithTag(text: String) {
+    onNodeWithTag(text).assertIsDisplayed().performClick()
+}
+
+internal fun inputTextWithTag(tag: String, text: String) {
+    onNodeWithTag(tag).performTextInput(text)
+}
+
+internal fun verifyRowsWithTextAreDisplayed(textList: List<String>) {
```
elihart, line 29: you could potentially make the invocation simpler by making these vararg instead

showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTestFlows.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+internal fun verifyRowsWithSubstringDisplayed(textList: List<String>) {
+    textList.forEach {
+        onNodeWithSubstring(it).assertIsDisplayed()
+    }
+}
+
+internal fun goBack(composeTestRule: AndroidComposeTestRule<ShowkaseBrowserActivity>) {
```
elihart, line 47: this might work nicely as an extension function

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
elihart: From what I can tell compose or the ui utils is not a dependency of this processor (not in build.gradle), which means you can't access it as a class. It will be dependent on whether the thing you are processing has that class present, which is why it is correct to access it in a nullable way, as you have to support any case of your processor being run in whatever environment, and can't make assumption about what will be present (or you have to add the dependency to the processor's build.gradle)

showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTest.kt
```diff
@@ -0,0 +1,392 @@
...
+        verifyRowsWithTextDoesNotExist(listOf("H1", "H2", "H3", "H4", "H5", "H6", "Subtitle1", 
+            "Subtitle2"))
+    }
+
+    @Test
+    fun navigating_to_component_leaf_screen_and_back_works_ok() {
+        // Ensure all the categories are visible
+        verifyLandingScreen()
+
+        // Select components to go to the component groups screen
```
elihart, line 311: while all these comments overall make the tests easy to understand, they are often redundant as the function names are very readable. redundant comments have the downside of 1. making twice as much text to read to understand the code 2. adding extra maintenance overhead and chance of comments getting out of sync with code

Generally I would recommend only adding comments that add additional information not available from the function invocation (although I wouldn't necessarily recommend changing all these as they are already done and it is quite nice overall)

vinaygaba: I'll keep this in mind for my future PR's 👍

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
vinaygaba: :O For some reason I remembered having all compose dependencies for showkase_testing. I might've cleaned it up at some point :O Should've just double checked. I think this is in fact doing the right thing 🥇

PR #105, opened 2020-09-08
Added support for passing @PreviewParameter to @Composable functions

vinaygaba:
This PR introduces the change needed to add support for passing `@PreviewParameter` to `@Composable` functions. Jetpack Compose allows passing a single `@PreviewParameter` parameter to a `@Preview` function. This PR adds the same functionality to `@Preview` & `@ShowkaseComposable` functions. 

`PreviewParameterProvider` returns a sequence of values and a new `Composable` function is generated for each value in the sequence. 

~I'll add unit tests to this PR once we agree with the general direction.~ Added unit tests

@airbnb/showkase-maintainers

### DIFF
--- a/sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
+++ b/sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.preferredWidth
 import androidx.compose.foundation.shape.RoundedCornerShape
@@ -104,16 +105,16 @@ fun CreditCardComponentPreview() {
 }
 
 
-data class DemoDataClass(
+data class BankHeader(
     val name: String,
     val age: Int
 )
 
-class NewParameterProvider : PreviewParameterProvider<DemoDataClass> {
-    override val values: Sequence<DemoDataClass>
+class NewParameterProvider : PreviewParameterProvider<BankHeader> {
+    override val values: Sequence<BankHeader>
         get() = sequenceOf(
-            DemoDataClass("John", 12),
-            DemoDataClass("Doe", 20)
+            BankHeader("Citi", 12),
+            BankHeader("Goldman Sachs", 20)
         )
 
     override val count: Int
@@ -121,10 +122,18 @@ class NewParameterProvider : PreviewParameterProvider<DemoDataClass> {
 
 }
 
-@Preview
+@Preview(group = "Payments")
 @Composable
-fun DemoComposablePreview(
-    @PreviewParameter(provider = NewParameterProvider::class) demoDataClass: DemoDataClass
+fun BankHeaderPreview(
+    @PreviewParameter(provider = NewParameterProvider::class) bankHeader: BankHeader
 ) {
-    Text(demoDataClass.name)
+    MaterialTheme {
+        Card {
+            Text(
+                text = bankHeader.name,
+                modifier = Modifier.fillMaxWidth().padding(padding4x),
+                style = MaterialTheme.typography.h2
+            )
+        }
+    }
 }
--- a/sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
+++ b/sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
@@ -14,7 +14,6 @@ import androidx.compose.foundation.text.CoreTextField
 import androidx.compose.runtime.Composable
 import androidx.compose.material.Card
 import androidx.compose.material.MaterialTheme
-import androidx.compose.runtime.state
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -32,6 +31,8 @@ import androidx.compose.ui.text.input.VisualTransformation
 import androidx.compose.ui.unit.dp
 import androidx.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.sp
+import androidx.ui.tooling.preview.PreviewParameter
+import androidx.ui.tooling.preview.PreviewParameterProvider
 import com.airbnb.android.showkase.annotation.ShowkaseComposable
 import com.airbnb.android.showkase.ui.padding4x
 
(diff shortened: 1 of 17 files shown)


sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
```diff
@@ -101,3 +102,29 @@ val creditCardOffsetMap = object : OffsetMap {
...
     CreditCardComponent("John Doe")
 }
+
+
+data class DemoDataClass(
+    val name: String,
+    val age: Int
+)
+
+class NewParameterProvider : PreviewParameterProvider<DemoDataClass> {
```
elihart, line 112: I see some opportunity to ship a small class that wraps this boilerplate with a basic list implementation (having to provide the count is annoying...). potentially a class that takes the list of examples in the constructor
```
class NewParameterProvider : PreviewParameterList {
 listOf(
            DemoDataClass("John", 12),
            DemoDataClass("Doe", 20)
 )
}
```

i guess we still want it to be lazy, so they aren't instantiated on each function call, so the constructor can take a lambda that returns a list

showkase-annotation/src/main/java/com/airbnb/android/showkase/annotation/ShowkaseCodegenMetadata.kt
```diff
@@ -34,5 +34,6 @@ annotation class ShowkaseCodegenMetadata(
     val showkaseWidthDp: Int = -1,
     val showkaseHeightDp: Int = -1,
     val insideWrapperClass: Boolean = false,
-    val insideObject: Boolean = false
+    val insideObject: Boolean = false,
+    val previewParameterClass: Array<KClass<*>> = [],
```
elihart, line 38: can't this be a an array of `PreviewParameter`? it seems like it should be able to be more specific (assuming the annotation processor has the compose library available at runtime as a dependency

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -153,6 +157,22 @@ internal fun ShowkaseCodegenMetadata.toModel(element: Element): ShowkaseMetadata
...
 
+private fun ShowkaseCodegenMetadata.getCodegenMetadataTypeMirror(): Pair<TypeMirror?, TypeMirror?> {
+    val enclosingClassTypeMirror = try {
+        enclosingClass
+        listOf<TypeMirror>()
+    } catch (mte: MirroredTypesException) {
+        mte.typeMirrors
+    }.firstOrNull()
+    val previewParameterClassTypeMirror = try {
+        previewParameterClass
```
elihart, line 168: to help yourself out i'd recommend creating an extension function to get a type mirror

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -261,10 +278,49 @@ internal fun getShowkaseMetadataFromPreview(
...
+private fun ExecutableElement.processPreviewParameterAnnotation(
+    typeUtils: Types,
+    previewParameterTypeMirror: TypeMirror
+) = parameters
+    .map {
+        val map = mutableMapOf<PreviewParameterProperty, Any>()
+        val previewParameterAnnotation = it.annotationMirrors.find {
+            typeUtils.isSameType(it.annotationType, previewParameterTypeMirror)
+        }
+        previewParameterAnnotation?.elementValues?.map { entry ->
```
elihart, line 310: I think you can use one of the `associate` std lib functions to avoid manually managing a mutable map? (could filter PreviewParameterProperty.values() first).

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -261,10 +278,49 @@ internal fun getShowkaseMetadataFromPreview(
...
+            val key = entry.key.simpleName.toString().toUpperCase(Locale.getDefault())
```
elihart, line 311: it's hard to understand what key and value really represent here. could they be named to better indicate what type they are?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
```diff
@@ -56,50 +56,51 @@ internal class ShowkaseBrowserWriter(private val processingEnv: ProcessingEnviro
             getPropertyList(SHOWKASE_BROWSER_COMPONENT_CLASS_NAME, COMPONENT_PROPERTY_NAME)
 
         val componentListInitializerCodeBlock =
-            SHOWKASE_BROWSER_COMPONENT_CLASS_NAME.listInitializerCodeBlock()
+            SHOWKASE_BROWSER_COMPONENT_CLASS_NAME.mutableListInitializerCodeBlock()
+        val showkaseMetadataWithoutParameterList = showkaseMetadataSet.filter { 
+            it is ShowkaseMetadata.Component && it.previewParameter == null
```
elihart, line 61: use `partition` for this I think
```
val (showkaseMetadataWithParameterList, showkaseMetadataWithoutParameterList) = showkaseMetadataSet.instanceOf<ShowkaseMetadata.Component>().partition { it.previewParameter != null }

elihart (approved):
@vinaygaba this direction seems pretty good, will be great to support preview data

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
vinaygaba: oh I initially did that but it crashes as even accessing that field throws the MirroredTypeException. So I reverted that change - https://github.com/airbnb/Showkase/pull/105/commits/e7721cc65c44ce18d8a314fb9cec3c3ef318a1d9 @elihart

showkase-annotation/src/main/java/com/airbnb/android/showkase/annotation/ShowkaseCodegenMetadata.kt
vinaygaba: the annotation processor does not have the compose library available at runtime. This was mainly due to how it requires a custom compiler and hence it wasn't possible to use compose without adding android dependencies. This might change in the future though or might've already change now @elihart

sample-submodule/src/main/java/com/airbnb/android/submodule/Payments.kt
vinaygaba: providing the count is actually optional so I can just remove that line. Also Preview would not know how to instantiate this class with data so passing parameters wouldn't be possible I believe @elihart

PR #122, opened 2020-12-09
Upgraded Showkase to the latest and greatest - alpha08

vinaygaba:
Kotlin version upgraded to 1.4.20
Compose upgraded to alpha08
Compose navigation upgraded to alpha03
Compose Tooling & Compsoe UI Test package coordinates upgraded
All newly introduced warnings cleaned up
Deprecated API's updated to recommended API's

@elihart @BenSchwab @airbnb/showkase-maintainers


build.gradle
```diff
@@ -4,15 +4,15 @@ buildscript {
...
+            'compose': '1.0.0-alpha08',
+            'composeNavigation': '1.0.0-alpha03',
             'detekt': '1.7.4',
             'espresso': '3.2.0',
             'gradle': '4.2.0-alpha15',
             'junit' : '4.12',
             'junitImplementation' : '1.1.1',
-            'kotlin': '1.4.0',
-            'kotlinCompilerVersion': '1.4.0',
+            'kotlin': '1.4.20',
```
elihart, line 14: 1.4.21 is out now

showkase-processor-testing/src/androidTest/java/com/airbnb/android/showkase_processor_testing/ShowkaseBrowserTest.kt
```diff
@@ -24,12 +24,20 @@ class ShowcaseBrowserTest {
...
                 createShowkaseBrowserIntent(
                     InstrumentationRegistry.getInstrumentation().targetContext
                 )
-            )
+            ),
+            activityProvider = { rule ->
+                var activity: ShowkaseBrowserActivity? = null
+                rule.scenario.onActivity { activity = it }
+                if (activity == null) {
+                    throw IllegalStateException("Activity was not set in the ActivityScenarioRule!")
```
elihart, line 37: a bit shorter
```suggestion
                    error("Activity was not set in the ActivityScenarioRule!")
```

or
```
return activity ?: error("Activity was not set in the ActivityScenarioRule!")
```
even shorter and removes the need to `!!` which is ideal

elihart (approved):

build.gradle
vinaygaba: Don't think the dependencies that compose is using have started using that so it might not work. I'll still give it a shot but don't expect it to work.

PR #125, opened 2021-02-02
Upgraded compose to alpha11

vinaygaba:
Upgraded Showkase to use the latest and greatest compose dependencies
Updated corresponding API changes that come with it.

@elihart @BenSchwab


elihart (approved):
Great :)

PR #130, opened 2021-02-28
Upgraded showkase to beta-01

vinaygaba:
- Did some cleanup
- Added new dependencies
- Made API changes that were required by the upgrade

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ buildscript {
             'androidXTestRules':'1.1.0',
             'assertJ': '3.16.1',
             'compose': '1.0.0-beta01',
+            'composeActivity': '1.3.0-alpha03',
             'composeConstraintLayout': '1.0.0-alpha03',
             'composeNavigation': '1.0.0-alpha08',
             'detekt': '1.7.4',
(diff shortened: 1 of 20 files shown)


build.gradle
```diff
@@ -4,15 +4,17 @@ buildscript {
...
+            'composeNavigation': '1.0.0-alpha08',
             'detekt': '1.7.4',
             'espresso': '3.2.0',
             'gradle': '4.2.0-alpha15',
             'junit' : '4.12',
             'junitImplementation' : '1.1.1',
-            'kotlin': '1.4.21-2',
-            'kotlinCompilerVersion': '1.4.21-2',
+            'kotlin': '1.4.30',
+            'kotlinCompilerVersion': '1.4.30',
```
elihart, line 17: 1.4.31 is out now

elihart (approved):
cool :)

You might want to update gradle too. latest is 6.8.3

build.gradle
vinaygaba: That's the one I'm using on AS. However, it seems like that version doesn't work with Compose just yet - https://kotlinlang.slack.com/archives/CJLTWPH7S/p1614538297223300?thread_ts=1614518656.164200&cid=CJLTWPH7S

PR #132, opened 2021-03-04
Changing rt lib for java 11 and adding java 11 to github CI

meggamind:
@vinaygaba 

- Seems like tools.jar is actually no longer needed with > Java 9.

### Issues with rt.jar
- The class and resource files previously stored in lib/rt.jar, lib/tools.jar, lib/dt.jar, and various other internal jar files will now be stored in a more efficient format in implementation-specific files in the lib directory. [StackOverflow](https://stackoverflow.com/questions/53707666/how-to-get-tools-jar-for-openjdk-11-on-windows)
- I greped and I found that abstractProcessor is in jmods/java.compiler.jmod
- However the JDK which comes with Android Studio doesn't have jmods, but Oracle JDK contains it.
- If you just copy this folder to android studio jdk path, it'll start working locally by adding `File runtimePath = new File(javaBase, "jmods/java.compiler.jmod")` to build.gradle
- I tried adding this jmod file in libs folder in the project, but it doesn't work the same way as rt.jar in libs folder.
 1. We can ask the users to move jmods folder to the android studio jdk directory and then use , however that obv is unreasonable.
 2. Ask users to download oracle jdk and then change the JAVA_HOME to that jdk and then we can use jmods
3. **Only working option as of now is to package the rt.jar in the libs folder.**

### About size
- It adds 64 MB, but I'll install git-lfs in the next PR which will reduce the size to clone and then pull the large file rt.jar later
`

### DIFF
--- a/showkase-processor-testing/build.gradle
+++ b/showkase-processor-testing/build.gradle
@@ -1,5 +1,3 @@
-import org.gradle.internal.jvm.Jvm
-
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
@@ -78,20 +76,6 @@ dependencies {
     androidTestImplementation deps.test.androidxTestRunner
     // Allows this module to access the annotation processor related classes. Otherwise those are
     // only available in java library modules.
-    testImplementation files(getRuntimeJar())
+    testImplementation files('libs/rt.jar')
     testImplementation deps.kotlinCompileTesting
-}
-
-static def getRuntimeJar() {
-    try {
-        final File javaBase = new File(System.getProperty("java.home")).getCanonicalFile()
-        File runtimeJar = new File(javaBase, "jmods/java.base.jmod")
-        if (runtimeJar.exists()) {
-            return runtimeJar
-        }
-        runtimeJar = new File(javaBase, "jre/jmods/java.base.jmod")
-        return runtimeJar.exists() ? runtimeJar : null
-    } catch (IOException e) {
-        throw new RuntimeException(e)
-    }
 }
--- a/showkase-processor-testing/build.gradle
+++ b/showkase-processor-testing/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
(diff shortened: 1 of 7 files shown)


showkase/build.gradle
```diff
@@ -21,11 +21,11 @@ kapt {
 }
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
 
     defaultConfig {
-        minSdkVersion 21
-        targetSdkVersion 29
+        minSdkVersion 26
```
vinaygaba, line 27: I know this is still work in progress and you added this only because of this build error

```
/Users/runner/work/Showkase/Showkase/showkase-processor-testing/libs/rt.jar: D8: com.android.tools.r8.internal.h1: MethodHandle.invoke and MethodHandle.invokeExact are only supported starting with Android O (--min-api 26)
```

But I'm keen on not updating this minSdkVersion. I think it's important to keep the library accessible and 21 is a reasonable `minSdkVersion`. There's still a way to use this if the consumers of this library add an override using `tools:overrideLibrary` but I would like to not have them do that.

showkase-processor-testing/build.gradle
```diff
@@ -76,24 +74,8 @@ dependencies {
...
     androidTestImplementation deps.test.androidXTestRules
     androidTestImplementation deps.test.androidxTestRunner
-    // Allows this module to access the annotation processor related classes. Otherwise those are 
-    // only available in java library modules. Inspiration - 
-    // https://github.com/airbnb/epoxy/blob/master/epoxy-processortest/build.gradle
-    testImplementation files(getRuntimeJar())
-    testImplementation files(Jvm.current().getToolsJar())
+    // Allows this module to access the annotation processor related classes. Otherwise those are
+    // only available in java library modules.
+    testImplementation files('libs/rt.jar')
```
vinaygaba, line 79: :( I was hoping we don't have to use this option of adding a Java 8 jar file separately in a Java 11 project

meggamind: yea there is no other option as of now. we can create a ticket for Android group and ask them to add jmods folder in jdk provided by android studio

meggamind: Orcale jdk does provide it.

vinaygaba: ah I see, interesting! Better than the Java8 option I guess.

PR #134, opened 2021-03-07
 Added the total count of UI elements in a given category and group

vinaygaba:
- Added the total count of UI elements in a given category and under a group. This makes it easy to see how many components, colors and text styles existing in a given codebase.
- Refactored some code so that it can be shared. 

Screen Shot 2021-03-06 at 7 55 53 PM

@airbnb/showkase-maintainers


showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -308,6 +314,24 @@ internal fun ShowkaseBodyContent(
...
+    val typographyCount = groupedTypographyMap.values.flatCount()
+    
+    return mapOf(
+        ShowkaseCategory.COMPONENTS to componentCount,
+        ShowkaseCategory.COLORS to colorsCount,
+        ShowkaseCategory.TYPOGRAPHY to typographyCount
+    )
+}
+
+internal fun Collection<List<*>>.flatCount() = flatMap { it }.count()
```
elihart, line 333: since this is so specialized it is unlikely to get used elsewhere anyway. Might as well make the receiver `Map<String, List<*>>` so you don't have to repeat the `values` part, and then make it a local function

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -308,6 +314,24 @@ internal fun ShowkaseBodyContent(
...
+    groupedComponentMap: Map<String, List<ShowkaseBrowserComponent>>,
+    groupedColorsMap: Map<String, List<ShowkaseBrowserColor>>,
+    groupedTypographyMap: Map<String, List<ShowkaseBrowserTypography>>,
+): Map<ShowkaseCategory, Int> {
+    val componentCount = groupedComponentMap.values.flatCount()
+    val colorsCount = groupedColorsMap.values.flatCount()
+    val typographyCount = groupedTypographyMap.values.flatCount()
+    
+    return mapOf(
+        ShowkaseCategory.COMPONENTS to componentCount,
```
elihart, line 327: imo removing the intermediate property is more readable `ShowkaseCategory.COMPONENTS to groupedComponentMap.flatCount()`

Otherwise it is an extra step to look at what the property name is and where it is used, introducing a chance of mistakenly using it in the wrong place

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseGroupsScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+    showkaseBrowserScreenMetadata: MutableState<ShowkaseBrowserScreenMetadata>,
+    navController: NavHostController,
+    onClick: () -> Unit
+) {
+    val filteredMap = getFilteredSearchList(
+        groupedTypographyMap.toSortedMap(),
+        showkaseBrowserScreenMetadata
+    )
+
+    val string = filteredMap.map { it.key }.joinToString()
```
elihart, line 28: where is string used? a better name would be easier to follow

elihart (approved):

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseGroupsScreen.kt
vinaygaba: not used anywhere..

PR #137, opened 2021-03-12
Fix version typo in README.md

adrinieto:


vinaygaba (approved):
Thanks for contributing! Good catch!

PR #140, opened 2021-03-17
Provide hooks to access Showkase metadata

vinaygaba:
@zach-klippenstein @laurakelly reached out with an interesting use case as they were discussing the possibility of using Showkase at Square - can we somehow use all the metadata that Showkase already has but stylize it and add our own skin to the browser. The main reason they wanted to do this was for consistency with the component browser they had for iOS. 

I thought this was a super valid use case and having hooks in place to access all the metadata Showkase has is actually needed for some of the things I've been wanting to do with Showkase(things like screenshot testing would also benefit from getting access to this metadata) :D 

To that effect, I made some changes to make this metadata accessible. Previously, the `createShowkaseBrowserIntent` method that created the intent for accessing the browser was generated as a top level function in a file. I've modified that file to generate an object called `ShowkaseMethods` that's going to generate static helper functions related to Showkase. Having it inside an object makes it easier to discover them. 

In addition, I now generate an additional function called `getShowkaseMetadata`. This function uses identical logic that the `ShowkaseBrowserActivity` uses for accessing the metadata using reflection - https://github.com/airbnb/Showkase/blob/master/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserActivity.kt#L57. 

This is what the generated object looks like

```kotlin
fun Showkase.createShowkaseBrowserIntent(context: Context): Intent {
    val intent = Intent(context, ShowkaseBrowserActivity::class.java)
    intent.putExtra("SHOWKASE_ROOT_MODULE", "com.airbnb.android.showkasesample.RootModule")
    return intent
}

fun Showkase.getShowkaseMetadata(): ShowkaseElementsMetadata {
    try {
      val showkaseComponentProvider =
          Class.forName("com.airbnb.android.showkasesample.RootModuleCodegen").newInstance() as
          ShowkaseProvider
      return showkaseComponentProvider.metadata()
    } catch(exception: ClassNotFoundException) {
      error("The class wasn't generated correctly. Make sure that you have setup Showkase correctly by following the steps here - https://github.com/airbnb/Showkase#Installation.")
    }
}
```

For anyone that wants to get access to the metadata, all they need to use is `Showkase.getShowkaseMetadata()`. The data class that this method returns has a maps of components, colors and typography grouped by their group names. 

Rest of the stuff is just clean up to support this change. 

@airbnb/showkase-maintainers @elihart @BenSchwab


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseMethodsWriter.kt
```diff
@@ -0,0 +1,154 @@
...
+        addOriginatingElement(rootElement)
+    }
+        .build()
+
+    private fun CodeBlock.Builder.generateShowkaseElementProperty(
+        variableName: String,
+        elementMethodName: String,
+        groupPropertyName: String
+    ) = addStatement(
+        "val $variableName = %L.$elementMethodName().groupBy { it.$groupPropertyName }",
```
vinaygaba, line 131: I'm actually going to update this and simply return a `Set` instead of a map that has the group name as key. I think returning the flat list should be good enough and I should not imply my preference by grouping this as it can be inferred from the metadata anyway. What do y'all think?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseMethodsWriter.kt
```diff
@@ -0,0 +1,154 @@
...
+        rootModuleClassName: String,
+        rootElement: Element
+    ) {
+        getFileBuilder(
+            rootModulePackageName,
+            "${rootModuleClassName}$SHOWKASE_METHODS_SUFFIX"
+        )
+            .addType(
+                TypeSpec.objectBuilder(SHOWKASE_METHODS_OBJECT_NAME)
+                    .addFunction(
```
vinaygaba, line 25: Will add kdoc for both the generated functions as well once we agree with...

laurakelly:
Wow, fast turnaround time! Thanks for adding this, I'll post to my team channel so we can discuss 😄

zach-klippenstein:
This is awesome, thanks!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseMethodsWriter.kt
```diff
@@ -0,0 +1,154 @@
...
+                )
+                .addStatement(
+                    "return %T(%L, %L, %L)",
+                    SHOWKASE_ELEMENTS_METADATA_CLASS_NAME,
+                    "componentsMap",
+                    "colorsMap",
+                    "typographyMap"
+                )
+                .unindent()
+                .addStatement("} catch(exception: ClassNotFoundException) {")
```
BenSchwab, line 114: Two questions: Why would a CNF exception happen? And is an metadata holder the right default?

BenSchwab:
Can we move any of this generated code to the runtime (a method on the interface?) as it seems like most of this code never change?

```
 fun getShowkaseElementsMetadata(): ShowkaseElementsMetadata {
      try {
        val showkaseComponentProvider =
            Class.forName("com.airbnb.android.showkasesample.RootModuleCodegen").newInstance() as ShowkaseProvider
        val componentsMap = showkaseComponentProvider.getShowkaseComponents().groupBy { it.group }
        val colorsMap = showkaseComponentProvider.getShowkaseColors().groupBy { it.colorGroup }
        val typographyMap = showkaseComponentProvider.getShowkaseTypography().groupBy { it.typographyGroup }
        return ShowkaseElementsMetadata(componentsMap, colorsMap, typographyMap)
      } catch(exception: ClassNotFoundException) {
        return ShowkaseElementsMetadata()
      }
  }
}

```

could just be

```
fun getShowkaseElementsMetadata(): ShowkaseElementsMetadata {
     val provider = Class.forName("com.airbnb.android.showkasesample.RootModuleCodegen").newInstance() as ShowkaseProvider
       return provider.metadata()
  ```

Also if we did this, would we even need the generated function? Couldn't we just do `RootModuleCodegen().metadata()`

vinaygaba:
@BenSchwab I think having the method as part of the interface and thus on `RootModuleCodegen` could work. It won't save us anything as the name of this file is dynamic so the user won't have an easy way to invoke it. That's part of the reason why we also generate the intent so that the name of the file can be inferred for the user. This logic has gone through a few iterations and that's probably why its dynamic but I do think that we can potentially keep this file name static.

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseMethodsWriter.kt
vinaygaba: I think notifying the user is a reasonable thing to do for this function.

elihart (commented):
@vinaygaba this seems like a great direction overall.

For high level feedback, I think it may be preferrable to have the object be part of the main library, and to generate extension functions for it. Referencing generated code can be a pain for discoverability and when the build hasn't been run yet, so this may help people. Also, if the main object is checked in to the library then you can easily add hardcode functions like Ben suggested (I like his suggestion about generating less code where possible).

Also, in terms of naming, consider naming the object `Showkase` instead of `ShowkaseMethods`. "Methods" is a bit redundant, since generally methods are what go inside objects, and shorter is better when no meaning is loss. 

you can also change the method names to remove "Showkase". `ShowkaseMethods. createShowkaseBrowserIntent` is redundant - in `createShowkaseBrowserIntent` we already have the Showkase content from the object name, so we can shorten it to `Showkase. createBrowserIntent` to provide a nicer API.

vinaygaba:
Did another pass and I think I understood what you meant. Thanks for the suggestion! @BenSchwab

BenSchwab (approved):

PR #145, opened 2021-04-14
Upgraded Showkase to Java11

vinaygaba:
Took another stab at upgrading Showkase to Java 11. Doing this required making changes to the dependency on `rt.jar` as that jar file is not available on in Java 11 anymore. Took the` rt.jar` file from the embedded Java files that comes with Android Studio and added it to the project.  Since its only for the testing module, this should be fine to do (although I do wish there was a better solution as this isn't my first preference). 

This PR is almost identical to where @meggamind landed in his PR as well - https://github.com/airbnb/Showkase/pull/132 with the exception that I'm only bumping up the `minSdkVersion` in the testing module and not the main module. This was a pretty important requirement for me personally as Showkase should be available to as many projects as possible so I didn't want to bump up the `minSdkVersion` for the entire project. I'm happy to either merge this PR or clean up the existing one and merge that PR!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
             'composeNavigation': '1.0.0-alpha09',
             'detekt': '1.7.4',
             'espresso': '3.2.0',
-            'gradle': '4.2.0-alpha15',
+            'gradle': '7.0.0-alpha12',
             'junit' : '4.12',
             'junitImplementation' : '1.1.1',
             'kotlin': '1.4.31',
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
             'composeNavigation': '1.0.0-alpha09',
             'detekt': '1.7.4',
             'espresso': '3.2.0',
-            'gradle': '4.2.0-alpha15',
+            'gradle': '7.0.0-alpha12',
```
elihart, line 13: nit: to me at least it's confusing to name this `gradle` since it's not exactly the gradle version. the android gradle plugin (agp) is different from gradle

elihart (approved):
🙏

PR #146, opened 2021-04-19
Support for KSP

elihart:
Work in progress. The large majority of the work is done, but several key lines can't be implemented yet; waiting for a few things in the abstraction library to be implemented:
https://github.com/androidx/androidx/pull/163
https://github.com/androidx/androidx/pull/164

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,6 +110,7 @@ allprojects {
         kotlinOptions {
             jvmTarget = "1.8"
             freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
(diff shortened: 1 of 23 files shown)


vinaygaba:
Exciting!! Going to brush up on KSP and start taking a look at the PR 🕺

elihart:
@vinaygaba Great :) Note that I'm using the XProcessing library from Androidx/Room (https://github.com/androidx/androidx/tree/androidx-main/room/compiler-processing), which enables kapt and ksp in the same code. Because of that I don't directly access any KSP api's

PR #150, opened 2021-05-26
Upgraded compose to beta07

nathan3d:
Updated Gradle distribution URL to use Gradle 7
Added Maven Central dependency (Removing JCenter requires updating Detekt)
Updated Gradle Maven Publish Plugin to 0.15.1
Fixed current route argument in Showkase Browser to match updated Compose Navigation API


nathan3d:
Looks like the included version of Detekt still needs JCenter, so added it back.

vinaygaba (approved):
Thank you for your contribution!

Issue #167, opened 2021-09-01
Crash when using MDC Compose Theme Adapter


hedvigoscar:
When using MDC Compose Theme Adapter (https://material-components.github.io/material-components-android-compose-theme-adapter/), and using `MdcTheme` in a `@Preview`, `ShowkaseBrowserActivity` crashes as the activity theme does not extend `Theme.MaterialComponents` (crash occurs on the following line: https://github.com/material-components/material-components-android-compose-theme-adapter/blob/774fe521be155b3a4c342394462965bb7aa3a9d8/lib/src/main/java/com/google/android/material/composethemeadapter/MdcTheme.kt#L168).

I attempted to work around the issue by overriding the theme declaration for `ShowkaseBrowserActivity`s theme:

```xml
<style name="Theme.App.NoActionBar" parent="Theme.MaterialComponents.DayNight.NoActionBar" />
```

which partially solved the issue - `ShowkaseBrowserActivity` no longer crashes on rendering a `@Composable` in the `ShowkaseComponentsInAGroupScreen`. The crash still happens on `ShowkaseComponentDetailScreen`, however, for reasons that are not entirely clear to me. I can only assume that it has to do with the `Context`-manipulation used for the different preview-flavors has something to do with it.

hedvigoscar:
>I can only assume that it has to do with the Context-manipulation used for the different preview-flavors has something to do with it.

This assumption appears to have been correct - the offending context is from `RTLComponentCard`. This appears, in turn, to be caused by `Context::createConfigurationContext` not inheriting the theme of its parent context, and hence upon resolution receiving the default theme for the platform (for API 29, where I'm performing these tests, this is `android:style/Theme.DeviceDefault.Light.DarkActionBar`).

I realize that working around this limitation might be out of scope for this library, but I think it might be possible to work around the issue by 'forwarding' the theme of the parent context to the newly created context by `Context::createConfigurationContext`. If time allows, I'll make a PR!

hedvigoscar:
Come to think of it - what is the reason that a configuration-context is created in this case? 🤔 

[...]

PR #168, opened 2021-09-02
Remove `Context`-manipulating code from `RTLComponentCard`

hedvigoscar:
Given `LocalLayoutDirection`, it is no longer necessary.

This should fix Issue #167.

### DIFF
--- a/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
+++ b/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
@@ -52,7 +52,6 @@ import com.airbnb.android.showkase.models.ShowkaseBrowserComponent
 import com.airbnb.android.showkase.models.ShowkaseBrowserScreenMetadata
 import com.airbnb.android.showkase.models.ShowkaseCurrentScreen
 import com.airbnb.android.showkase.models.update
-import java.util.Locale
 
 @Composable
 internal fun ShowkaseComponentDetailScreen(
--- a/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
+++ b/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
@@ -169,20 +169,12 @@ private fun DisplayScaledComponentCard(metadata: ShowkaseBrowserComponent) {
 
 @Composable
 private fun RTLComponentCard(metadata: ShowkaseBrowserComponent) {
-    val customConfiguration = Configuration(LocalConfiguration.current).apply {
-        val locale = Locale("ar")
-        setLocale(locale)
-        setLayoutDirection(locale)
-    }
-    val customContext = LocalContext.current.createConfigurationContext(customConfiguration)
     ComponentCardTitle("${metadata.componentName} [RTL]")
-    CompositionLocalProvider(LocalContext provides customContext) {
-        val updatedModifier = Modifier.generateComposableModifier(metadata)
-        Card(modifier = Modifier.fillMaxWidth()) {
-            CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Rtl) {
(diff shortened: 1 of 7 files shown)


showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
```diff
@@ -169,20 +168,12 @@ private fun DisplayScaledComponentCard(metadata: ShowkaseBrowserComponent) {
...
-    CompositionLocalProvider(LocalContext provides customContext) {
-        val updatedModifier = Modifier.generateComposableModifier(metadata)
-        Card(modifier = Modifier.fillMaxWidth()) {
-            CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Rtl) {
-                Column(modifier = updatedModifier) {
-                    metadata.component()
-                }
+    val updatedModifier = Modifier.generateComposableModifier(metadata)
+    Card(modifier = Modifier.fillMaxWidth()) {
+        CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Rtl) {
```
vinaygaba, line 174: I'm guessing you ran this locally and confirmed that the text was indeed in the RTL orientation after this change? @hedvigoscar

vinaygaba (commented):
Would be amazing if you could also add a test case for the crash you were encountering. There's a lot of test setup already in place. All you will need to do is to simply add a `@ShowkaseComposable` or `@Preview` that leverages `MDC Theme Adaptor` and add it in here - https://github.com/airbnb/Showkase/tree/master/showkase-processor-testing/src/main/java/com/airbnb/android/showkase_processor_testing

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
hedvigoscar: I did indeed. 🙂

hedvigoscar:
> Would be amazing if you could also add a test case for the crash you were encountering. There's a lot of test setup already in place. All you will need to do is to simply add a `@ShowkaseComposable` or `@Preview` that leverages `MDC Theme Adaptor` and add it in here - [`master`/showkase-processor-testing/src/main/java/com/airbnb/android/showkase_processor_testing](https://github.com/airbnb/Showkase/tree/master/showkase-processor-testing/src/main/java/com/airbnb/android/showkase_processor_testing)

Done!

The implications of adding `MdcTheme` in the integration tests is that the theme of `ShowkaseBrowserActivity` now needs to extend `Theme.MaterialComponents` - I leave it up to you whether this is a worthwhile change to make.

(technically, we might be able to override the theme in the testing infrastructure? I might investigate that, in order to make the library just slightly lighter on dependencies)

I added `MdcTheme` to one of the existing test composables, as the tests are tightly coupled to the amount of test composables present in the module. Running the tests with a theme not extending `Theme.MaterialComponents` results in a predictable crash.

hedvigoscar:
>(technically, we might be able to override the theme in the testing infrastructure? I might investigate that, in order to make the library just slightly lighter on dependencies)

I did this instead, and the dependency on `material` in the main library is no longer required. Nice!

vinaygaba (approved):
Thank you for finding this issue and fixing it! We appreciate your contributions!

hedvigoscar:
You're welcome - thanks for looking into merging it so soon. 🙂

Can we get this into a release soon as well? 😉 We're excited to continue integrating `Showkase` into our workflow, but we won't be able to until this is live 😬

vinaygaba:
@hedvigoscar I can get a release out first thing tomorrow morning! I was hoping to squeeze in the latest version of Compose but given its been a while since the last release, I can get one out tomorrow!!

hedvigoscar:
Lovely! Big thanks 😄

vinaygaba:
@hedvigoscar Just published the latest release - https://github.com/airbnb/Showkase/releases/tag/1.0.0-beta04

PR #174, opened 2021-09-22
Remove app_name string identifier

### LINKED ISSUE #173: app_name string is overridden by Showkase
The file [showkase/src/main/res/values/strings.xml](https://github.com/airbnb/Showkase/blob/master/showkase/src/main/res/values/strings.xml) contains a string with the key `app_name` and the value `Showkase`.

Through manifest merging, this ends up overwriting our app's name in the manifest file.

I don't see a case for this string to exist in Showkase at all 🤔

StefMa:
Closes #173


StefMa:
Closed in favor of #175

PR #175, opened 2021-09-22
Rename string app_name to showkase_title

### LINKED ISSUE #173: app_name string is overridden by Showkase

guk-ioki:
Closes #173


vinaygaba (approved):
Thank you for your contribution! I appreciate it!

PR #194, opened 2021-12-02
KSP Support

elihart:
Refactored the processor to use the XProcessing abstraction so that ksp can be supported as well. Processor tests were updated to run in both kapt and ksp mode, and all tests pass.

For the application modules, I updated the build.gradle files to switch between applying ksp vs kapt with a gradle property `-PuseKsp=true` (defaults to kapt).  Then I added a CI step to run connected checks for ksp.

I ran into a duplicate class build issue that took a while to resolve. while trying to fix that i updated a bunch of dependencies and cleaned some things up, and ultimately had to add an exclusion for the jetbrains annotations coming in two packages with different versions.

I also added basic timing in the processor, which is logged as a Note and can be viewed with the gradle -i flag. Ksp appears to be slightly quicker than kapt, which is great. note that ksp is actually much better overall because this doesn't capture the savings from not running the kapt generate stubs task, so about 70% faster overall.

```
//./gradlew sample:clean sample:kspDebugKotlin -i -PuseKsp=true
ksp ShowkaseProcessor [Round 1][ksp] finished in 284.46 ms

//  ./gradlew sample:clean sample:kaptDebugKotlin -i
kapt ShowkaseProcessor [Round 1][javac] finished in 360.427 ms
+
// ./gradlew sample:clean sample:kaptGenerateStubsDebugKotlin --profile
 -> 0.570s
```


showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_inside_companion_object_with_preview_annotation_compiles_ok/output/ShowkaseMetadataShowkase_processor_testing.kt
```diff
@@ -17,7 +17,9 @@ public class ShowkaseMetadataShowkase_processor_testing {
     showkaseKDoc = "",
     enclosingClass = [ShowkaseClass::class],
     showkaseMetadataType = "COMPONENT",
-    isDefaultStyle = false
+    isDefaultStyle = false,
+    showkaseWidthDp = -1,
```
elihart, line 21: xprocessing doesn't distinguish between default param values and custom values when retrieving values off an annotation, so these end up getting set, but the behavior is the same.

.github/workflows/android.yml
```diff
@@ -24,10 +24,8 @@ jobs:
         # https://medium.com/@ychescale9/running-android-instrumented-tests-on-ci-from-bitrise-io-to-github-actions-a611e57ce71f
         key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}
         restore-keys: ${{ runner.os }}-gradle-
-    - name: Build with Gradle
-      run: ./gradlew build
-    - name: Run Unit tests
-      run: ./gradlew test --stacktrace
+    - name: Run Checks
```
elihart, line 27: I don't think detekt was running before. "check" builds all modules, runs all tests, and runs detekt and other linters.

vinaygaba: I think it definitely was as I've had to fix multiple detekt errors in other PR's. Might be from some other task but it was running for sure

showkase-annotation/src/main/java/com/airbnb/android/showkase/annotation/ShowkaseComposable.kt
```diff
@@ -47,8 +47,8 @@ package com.airbnb.android.showkase.annotation
  * Use this to restrict the size of your preview inside the Showkase browser.
  */
 @MustBeDocumented
-@Retention(AnnotationRetention.RUNTIME)
-@Target(AnnotationTarget.FUNCTION, AnnotationTarget.TYPE)
```
vinaygaba, line 51: Trying to think if I added `TYPE` intentionally 🤔

showkase-browser-testing/build.gradle
```diff
@@ -43,20 +42,24 @@ android {
         exclude 'META-INF/AL2.0'
         exclude 'META-INF/LGPL2.1'
     }
+    configurations {
+        all {
```
vinaygaba, line 46: add a comment around why we are adding this otherwise there's no way we will remember it (like I've forgotten 20 other things throughout this codebase 😅)

showkase-processor-testing/build.gradle
```diff
@@ -43,23 +42,30 @@ android {
         exclude 'META-INF/AL2.0'
         exclude 'META-INF/LGPL2.1'
     }
+
+    configurations {
+        all {
+            // work around this error:
```
vinaygaba, line 48: nice!

showkase-processor-testing/build.gradle
```diff
@@ -75,9 +81,6 @@ dependencies {
     testImplementation deps.test.assertJ
     testImplementation deps.test.googleTruth
     testImplementation deps.test.junit
-    // Allows this module to access the annotation processor related classes. Otherwise those are
-    // only available in java library modules. Inspiration - 
-    // https://github.com/airbnb/epoxy/blob/master/epoxy-processortest/build.gradle
-    testImplementation files('libs/rt.jar')
```
vinaygaba, line 81: How were we able go get around needing this?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/BaseProcessor.kt
```diff
@@ -0,0 +1,118 @@
...
+        internalProcess(environment, XRoundEnv.create(environment, roundEnv))
+
+        if (roundEnv.processingOver()) {
+            finish()
+        }
+
+        return false
+    }
+
+    final override fun process(resolver: Resolver): List<KSAnnotated> {
```
vinaygaba, line 73: just for my own curiosity, when is this overloaded method called vs the other process method?

showkase-annotation/src/main/java/com/airbnb/android/showkase/annotation/ShowkaseComposable.kt
elihart: the processor in its current form would fail if any classes were annotated, so it shouldnt be there now at least

showkase-processor-testing/build.gradle
elihart: the jar was still needed but it was already being pulled in by `implementation fileTree(include: ['*.jar'], dir: 'libs')`

I actually reverted this and deleted the general import though since this is more explicit

vinaygaba (approved):
This is absolutely fantastic work 👏 Really excited to see the improvements with ksp, in addition to the clean up you did in this PR!

PR #195, opened 2021-12-02
Refactor processor test resources and update kotlin poet

elihart:
I needed to update kotlin poet, but the newer version has some slight changes to output, notably that explicit public modifiers are now added. It would have been quite tedious to manually update all of the hard coded inputs and outputs of the processor tests, so I took this opportunity to refactor them.

The new approach stores all inputs and outputs in resource folders, with a specific folder per test class and test method. The resources are automatically retrieved based on the test name so the actual test function declaration is essentially just stating the test name and whether it expects to succeed or fail with a certain message.

There is also a new boolean flag that can be enabled to automatically update the outputs, which I used for the kotlinpoet update. This is split into a separate commit for easy reviewing.


vinaygaba (approved):
What a fantastic improvement to Showkase 👏 

I'm guessing the actual input and output code does not have a lot of changes? I did go over a lot of it and nothing jumped out. Really excited about being able to make changes more easily to the tests which has historically been fairly painful to do. 

Just one nit - get rid of the committed .DS_Store files

PR #196, opened 2021-12-06
[Impression Logging] Added basic support to capture composable impressions

vinaygaba:
This PR introduces a new modifier that allows you to get visibility information of a given composable that uses this modifier. This will be helpful for things like impression logging that are very common in most apps these days. 

This doesn't yet accurate work with the LazyX composables but I plan to make those improvements in a follow up PR. 

**Usage** 

```kotlin
BasicText(
    text = "Testing Impression Events",
    modifier = Modifier
        .visibilityEvents(
            key = "key",
            onVisibilityChanged = { showkaseVisibilityEvent ->
               // Use the visibility event as per your use case
                Log.e(TAG, "key: ${showkaseVisibilityEvent.key} visibilityPercentage: ${showkaseVisibilityEvent.visibilityPercentage}")
            }
        )
)
```
@airbnb/showkase-maintainers


showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.collect
+
+/**
+ * Passed in the callback of the [visibilityEvents] Modifier. Contains information about the
+ * visibility of a given composable function that uses the visibilityEvents Modifier.
+ */
+data class ShowkaseVisibilityEvent<T>(
```
elihart, line 28: don't think a null key makes sense? also should you document what the key means/represents?
```suggestion
data class ShowkaseVisibilityEvent<T : Any>(
```

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+    val key: T,
+    val visibilityPercentage: Float,
+    val boundsInWindow: Rect
+)
+
+/**
+ * Use this modifier to get visibility events for a given Composable. It emits visibility events
+ * when the composable is added to the composition (visible), when its removed from the
+ * composition(invisible), when the activity is backgrounded(visible) and when the activity is
```
elihart, line 37:
```suggestion
 * composition(invisible), when the activity is backgrounded(invisible) and when the activity is
```

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+ * foregrounded(visible). In addition,
+ *
+ * @param key Unique identifier for a given composable function that you use this modifier on
+ * @param onVisibilityChanged Callback that's called when the visibility of a composable function
+ * changes. This event has information about the key of the composable, the percentage of composable
```
elihart, line 42: i think comments like `This event has information about the key of the composable, the percentage of composable
 * that's visible on the screen and the bounds of the composable.` can be omitted from here and instead should just be on that class

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+ */
+@OptIn(FlowPreview::class)
+fun <T> Modifier.visibilityEvents(
+    key: T,
+    onVisibilityChanged: (ShowkaseVisibilityEvent<T>) -> Unit,
+) = composed {
+    val view = LocalView.current
+    val lifecycle = LocalLifecycleOwner.current.lifecycle
+    val scope = rememberCoroutineScope()
+    val visibilityEvent by rememberUpdatedState(newValue = onVisibilityChanged)
```
elihart, line 53:
```suggestion
    val visibilityEventCallback by rememberUpdatedState(newValue = onVisibilityChanged)
```

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+/**
+ * Used for handling the use case where a composable function is not in composition anymore i.e is
+ * invisible.
+ */
+@SuppressLint("ComposableNaming")
+@Composable
+private fun <T> registerDisposeImpressionEvents(
```
elihart, line 73: what's the benefit of breaking this out from the original function? seems simple enough to keep in the main function. I also wonder if it is worth optimizing a hot path in library to minimize how many functions it calls out to to reduce recomposition work.

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+    key: T,
+    impressionCollector: ImpressionCollector<T>,
+    lifecycle: Lifecycle
+) {
+    DisposableEffect(key, lifecycle) {
+        onDispose {
+            impressionCollector.publishDisposeEvent(
+                key,
```
elihart, line 81: should be able to pass the key as a constructor parameter to impressionCollector instead of needing to pass it with every callback, since the key should always be the same

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/VisibilityUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import android.view.View
+import androidx.annotation.FloatRange
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.layout.LayoutCoordinates
+import androidx.compose.ui.layout.boundsInParent
+import androidx.compose.ui.layout.boundsInRoot
+import androidx.compose.ui.layout.boundsInWindow
+
+internal val defaultVisibilityMetadata = VisibilityMetadata(0f, Rect(0f, 0f,0f,0f))
```
elihart, line 11: maybe a better name is `empty` or `hidden` visibility metadata?

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+ * that's visible on the screen and the bounds of the composable.
+ */
+@OptIn(FlowPreview::class)
+fun <T> Modifier.visibilityEvents(
+    key: T,
+    onVisibilityChanged: (ShowkaseVisibilityEvent<T>) -> Unit,
+) = composed {
+    val view = LocalView.current
+    val lifecycle = LocalLifecycleOwner.current.lifecycle
+    val scope = rememberCoroutineScope()
```
elihart, line 52: maybe wrap this entire thing in a key block so you don't have to key each individual thing
```
key(key) {



}

File: showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt [hunk 9]
```diff
@@ -0,0 +1,145 @@
...
+ */
+@SuppressLint("ComposableNaming")
+@FlowPreview
+@Composable
+private fun <T> collectImpressionEvents(
+    key: T,
+    impressionCollector: ImpressionCollector<T>,
+    onVisibilityEvent: (ShowkaseVisibilityEvent<T>) -> Unit
+) {
+    LaunchedEffect(key) {
```
elihart, line 100: maybe the coroutine scope should be keyed instead of this effect, then you don't need to pass the key or worry about the scoping

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+        impressionCollector.impressionEvents
+            .collect { impressionData ->
```
elihart, line 102: it feels a bit weird to me to have the impression collector collect the events, but then expose them as a flow for this code to recollect to do the actual callback. Could you instantiate the impression collector with the coroutine scope and callback lambda and have that class manage all of this instead? that might make recomposition more efficient too

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+    key: T,
+    lifecycle: Lifecycle,
+    impressionCollector: ImpressionCollector<T>,
+    visibility: VisibilityMetadata?,
+) {
+    remember(key1 = key, key2 = visibility) {
+        lifecycle.addObserver(
+            LifecycleEventObserver { _, event ->
+                when (event) {
+                    Lifecycle.Event.ON_PAUSE, Lifecycle.Event.ON_STOP  -> {
```
elihart, line 130: I don't think it makes sense to use both resume/start, as checking both isn't symmetric - ie you will publish event in start and then in pause. pick either resume/pause or start/stop to be symmetric

elihart: i'm also not sure that it is ideal for the library to make the decision about which lifecycle events to use. some people might want created/destroyed, or they may not care about lifecycle changes and only want to know about percent visible.

maybe it would be better to add lifecycle state to the `ShowkaseVisibilityEvent` instead of internally using that information to set the visibility (in the spirit of being more generic and letting the user decide how they want to interpret the data)

the library can always add a util on top of that raw data as a default way to interpret it, eg an extension function like `ShowkaseVisibilityEvent.filterForLifecycle(Lifecycle.Event.ON_START)`

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
```diff
@@ -0,0 +1,55 @@
...
+        visibilityPercentage: VisibilityMetadata,
+    ) {
+        publishImpressionEvent(key, visibilityPercentage)
+    }
+
+    internal fun publishImpressionEvent(
+        key: T,
+        visibilityMetadata: VisibilityMetadata,
+    ) {
+        coroutineScope.launch {
```
elihart, line 32: I think right now this scope is not keyed so this will not be properly cancelled

build.gradle
```diff
@@ -32,6 +32,7 @@ buildscript {
             'strikt':'0.33.0',
```
elihart, line 32: I don't think I see the coroutines dependency here, where is that version being set?

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+ */
+@SuppressLint("RememberReturnType", "ComposableNaming")
+@Composable
+private fun <T> registerLifecycleImpressionEvents(
+    key: T,
+    lifecycle: Lifecycle,
+    impressionCollector: ImpressionCollector<T>,
+    visibility: VisibilityMetadata?,
+) {
+    remember(key1 = key, key2 = visibility) {
```
elihart, line 126: why does visibility need to be a key? won't this add a new observer each time visibility percentage changes?

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,145 @@
...
+        lifecycle.addObserver(
```
elihart, line 127: where is this observer removed? don't we need to remove it as a dispose effect? seems like this could be in the same `DisposableEffect(key, lifecycle)` block used above

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.launch
+
+internal class ImpressionCollector<T>(
+    val coroutineScope: CoroutineScope,
+    val onVisibilityEvent: (ShowkaseVisibilityEvent<T>) -> Unit
+) {
+    private val mutableImpressionEvents = MutableStateFlow<ShowkaseVisibilityEvent<T>?>(null)
+
+    @OptIn(FlowPreview::class)
+    val impressionEvents: Flow<ShowkaseVisibilityEvent<T>> =
```
elihart, line 18: the `debounce(200)` is an interesting decision to hide as an implementation detail. maybe it should be exposed as a parameter to the modifier and be optional?

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
```diff
@@ -0,0 +1,55 @@
...
+        mutableImpressionEvents.filterNotNull().debounce(200)
+
+    internal fun onLayoutCoordinatesChanged(
+        key: T,
+        visibilityPercentage: VisibilityMetadata,
+    ) {
+        publishImpressionEvent(key, visibilityPercentage)
+    }
+
+    internal fun publishImpressionEvent(
```
elihart, line 28: it's confusing that some of these "publish*" functions actually call back to `onVisibilityEvent`, and some emit to the flow, which triggers another call down the line. it is hard to follow

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
```diff
@@ -0,0 +1,55 @@
...
+                ShowkaseVisibilityEvent(
+                    key,
+                    visibilityMetadata.visibilityPercentage,
+                    visibilityMetadata.boundsInWindow
+                )
+            )
+        }
+    }
+
+    internal fun publishDisposeEvent(
```
elihart, line 43: consider using naming like `onDisposeEvent` - the caller tells this collector what happened, and publishing a callback is an implementation detail of the collector

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
vinaygaba: The unit of recompositions is a composable function/lambda so at best it would be more efficient and at worst it would have the same impact.

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
vinaygaba: I think LaunchedEffect guarantees that the scope is the "composition" so that shouldn't be a concern as it auto-cancels any ongoing coroutines.

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
vinaygaba: This was added because I noticed that onGlobalPositioned often publishes multiple identical events and so debouncing was important. I figured this didn't need to be exposed by the library and choosing a reasonable default was enough as its an implementation detail for the end user that they don't need to be aware about. Do you think there's a use case where having control over this would be valuable?

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
vinaygaba: Since I got this coroutine scope using `rememberCoroutineScope`, its tied to the scope of the composition and will be auto-cancelled as soon as this composable is not in composition.

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ImpressionCollector.kt
vinaygaba: I hear you and this was important to do to get around an issue that I spent a couple days debugging. Since I am using the coroutineScope that's tied to the composition (using `rememberCoroutineScope`), it causes the coroutine to be called. Thus, I cannot invoke a coroutine from the onDispose method and was not able to publish to the flow directly.

build.gradle
vinaygaba: The coroutine dependency is coming from compose itself that exposes it. I think it uses the version that works with the version of Kotlin that compose is built against.

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
vinaygaba: I think this pattern is followed in multiple places in the Compose codebase. This could've been a compose state object too but keeping it as a flow object gives me access to the operators that I was hoping to use.

vinaygaba: Wondering what you had in mind though. Could you elaborate a bit more @elihart

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Passed in the callback of the [visibilityEvents] Modifier. Contains information about the
+ * visibility of a given composable function that uses the visibilityEvents Modifier.
+ *
+ * @param key The unique identifier that you passed with the [visibilityEvents] modifier.
+ * @param visibilityPercentage A value between 0.0 and 1.0 that represents how much a given composable
+ * was visible on the screen. 0.0 represents that it wasn't visible on the screen and 1.0 represents
+ * that it was completely visible on the screen.
+ * @param boundsInWindow A [Rect] object that contains information about the bounds of a given
+ * composable. When the bounds are (0, 0, 0, 0), it means that the composable was not visible on
```
elihart, line 33: > When the bounds are (0, 0, 0, 0), it means that the composable was not visible on screen.

could use a sealed class instead to represent this in the type system more clearly

showkase-component-impressions/src/main/java/com/vinaygaba/showkase_component_impressions/ComponentImpressions.kt
```diff
@@ -0,0 +1,138 @@
...
+    val key: T,
+    val visibilityPercentage: Float,
+    val boundsInWindow: Rect,
+)
+
+/**
+ * Use this modifier to get visibility events for a given Composable. It emits visibility events
+ * when the composable is added to the composition (visible), when its removed from the
+ * composition(invisible), when the activity is backgrounded(invisible) and when the activity is
+ * foregrounded(visible). In addition,
```
elihart, line 46: dangling `In addition,`

elihart (approved):

PR #202, opened 2022-01-04
Added weight attr to Text in ShowkaseColorsInAGroupScreen

jaloveeye:
If the name was long(eg.  DarkOnSecondaryContainer) in the ShowkaseColor list, the color was not shown.
I added weight attr to text sovled [it.[](url)](url)


gradlew.bat
```diff
@@ -1,84 +1,84 @@
-@if "%DEBUG%" == "" @echo off
```
vinaygaba, line 1: Can you remove all the gradlew changes from this PR

vinaygaba (commented):
Thank you for contributing, I appreciate it!

Looks like there's a bunch of unintended changes in this PR (gradlew changes and changes to the test files). Can you remove those changes from the PR.

Additionally, could you share before/after screenshots so that the reviewer can visualize the changes 🙏

PR #208, opened 2022-01-15
Fix license URL

ofalvai:
I'm running [Licensee](https://github.com/cashapp/licensee) on my project. It flags this library with the following error, even though I allow Apache 2.0 licenses:

```
com.airbnb.android:showkase-annotation:1.0.0-beta10
 - ERROR: Unknown license URL 'https://github.com/airbnb/Showkase/blob/master/LICENSE' is NOT allowed
 ```

I think the license URL should point to the canonical URL, not to this GitHub repo


vinaygaba (approved):
Thank you for your contribution 🙏

Issue #210, opened 2022-01-17
Skip default group for typography


themanol:
Do you think makes sense to avoid showing the group intermediate screen in typography category when there is only one group?

I can create a PR for showing directly the list of typographies in case there is a single group if you think it is fine.

Thanks!

PR #211, opened 2022-01-17
ISSUE 210: skip default group for typography

themanol:
When there is only a group in typography skip the screen of selecting the group and show directly the typography list.

typography_single_group

https://github.com/airbnb/Showkase/issues/210


vinaygaba (commented):
Can you also add a test case in to test this - https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt

themanol:
I am not sure how I can add a test case here. 

The ShowkaseBrowserTest is generating the typography screens based on the TestTypography.kt file. Do you know how I can use a different Typography file for a specific test?

Thanks

vinaygaba:
Ah I see what you mean. I think it's fine for now in that case.

vinaygaba (approved):

Issue #212, opened 2022-01-18
Showkase was not able to find the root class that youpassed to @ShowkaseScreenshot


zoey-juan:
Hi!
I am trying to integrate the screenshot test but always getting `Showkase was not able to find the root class that youpassed to @ShowkaseScreenshot`. Not sure if I missed some configuration, could you please give me some hints? :)

./gradlew xxx-compose-app:runDebugAndroidTestScreenshotTest

```
dependencies {
    ...
    implementation(Libs.showkase)
    ksp(Libs.showkase_processor)
    androidTestImplementation("com.facebook.testing.screenshot:core:0.13.0")
    androidTestImplementation("com.airbnb.android:showkase-screenshot-testing:1.0.0-beta10")
    kaptAndroidTest("com.airbnb.android:showkase-processor:1.0.0-beta10")
}
```
```
@ShowkaseRoot
class RootModule : ShowkaseRootModule
```

```
@ShowkaseScreenshot(rootShowkaseClass = RootModule::class)
abstract class ComposeTests : ShowkaseScreenshotTest {

    override fun onScreenshot(
        id: String,
        name: String,
        group: String,
        styleName: String?,
        screenshotType: ShowkaseScreenshotType,
        screenshotBitmap: Bitmap
    ) { .....
    }
}
```

```
e: [ksp] com.airbnb.android.showkase.processor.exceptions.ShowkaseProcessorException: Showkase was not able to find the root class that youpassed to @ShowkaseScreenshot. Make sure that you have configured Showkase correctly.
        at com.airbnb.android.showkase.processor.ShowkaseProcessor.writeScreenshotTestFiles(ShowkaseProcessor.kt:291)
        at com.airbnb.android.showkase.processor.ShowkaseProcessor.processShowkaseMetadata(ShowkaseProcessor.kt:198)
        at com.airbnb.android.showkase.processor.ShowkaseProcessor.process(ShowkaseProcessor.kt:61)
        at com.airbnb.android.showkase.processor.BaseProcessor.internalProcess(BaseProcessor.kt:96)
        at com.airbnb.android.showkase.processor.BaseProcessor.process(BaseProcessor.kt:81)
        at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$4$1.invoke(KotlinSymbolProcessingExtension.kt:169)
[...]

PR #218, opened 2022-02-08
Fix unable to find showkase root issue

vinaygaba:
There were reports of this issue(https://github.com/airbnb/Showkase/issues/212) where screenshot tests weren't able to generate due to a build issue after the KSP upgrade. This was happening only if the `ShowkaseRoot` implementation was not declared in an `androidTest` sourceSet. Since the sample used in Showkase relied on a root implementation declared in the `androidTest` sourceSet, I wasn't able to repro this issue locally. I spent some time trying to figure out why the class wasn't being found on the classpath and wondered if it had something to do with XProcessing. Eventually, it struck me that we had also updated the retention of the annotation.

Thanks @davidvavra for sharing the project where I was able to observe what was causing the issue. 

@airbnb/showkase-maintainers


BenSchwab (approved):

elihart (approved):
good find, thanks!

PR #222, opened 2022-02-15
Issue-221 Skip categories screen in case of having only components

themanol:
showkase_components_only


marcorighini:
What do you think about passing this option as a flag in the showkase activity intent? @vinaygaba @themanol 
E.g. pass a configuration data class to the showkase activity intent that contains some configurations, i.e. atm `skipSingleScreens` (naming to be improved)

marcorighini:
@vinaygaba @themanol Do you think the suggestion above makes sense?

themanol:
> @themanol I've somehow missed seeing this PR. Truly apologize for that. I'm ready to merge it once the builds pass.

The build pass, we need a review here, Thanks! @vinaygaba

marcorighini:
Yes, looks like can be merged now. Would be great to have a new release with this.

vinaygaba:
> What do you think about passing this option as a flag in the showkase activity intent? @vinaygaba @themanol
E.g. pass a configuration data class to the showkase activity intent that contains some configurations, i.e. atm skipSingleScreens (naming to be improved)

@marcorighini I think it probably makes sense to have this as the default behavior instead of making it configurable. If a lot of think making it configurable is valuable, I'm open to reconsidering it.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -370,10 +370,74 @@ internal fun ShowkaseBodyContent(
     groupedTypographyMap: Map<String, List<ShowkaseBrowserTypography>>,
     showkaseBrowserScreenMetadata: MutableState<ShowkaseBrowserScreenMetadata>
 ) {
+    val isOnlyComponents = groupedColorsMap.values.isEmpty() && groupedTypographyMap.values.isEmpty()
     NavHost(
         navController = navController,
-        startDestination = ShowkaseCurrentScreen.SHOWKASE_CATEGORIES.name
-    ) {
+        startDestination = if (isOnlyComponents) {
```
vinaygaba, line 376: I think we should make this generic i.e directly jump to the respective groups screen if only one type of elements are present.

For example, if there are only colors, it should directly go to the colors groups screen. Same with typography or any other element that we might add support for in the future.

vinaygaba (commented):
Sorry about dropping on the ball on this PR and not reviewing it earlier.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
themanol: ok, it makes sense, I can change this to be the default behaviour

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseCategoriesScreen.kt
```diff
@@ -82,6 +83,9 @@ internal fun goBackToCategoriesScreen(
         showkaseBrowserScreenMetadata.value.isSearchActive -> {
             showkaseBrowserScreenMetadata.clearActiveSearch()
         }
+        navController.currentDestination?.id == navController.graph.startDestinationId -> {
+            activity.finish()
```
vinaygaba, line 87: I didn't follow this use case. Could you elaborate on which scenario we are trying to close the activity for?

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseCategoriesScreen.kt
```diff
@@ -74,6 +74,7 @@ private fun goBackFromCategoriesScreen(
 }
 
 internal fun goBackToCategoriesScreen(
+    activity: AppCompatActivity,
```
vinaygaba, line 77: Not a good idea to pass the activity. Instead you should pass the lambda

vinaygaba:
```suggestion
    onFinish:() -> Unit,
```

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseCategoriesScreen.kt
themanol: This happen because we go directly to the Component screen, so we are already on the root of the nav graph, so when we click back, we don't have to navigate to SHOWKASE_CATEGORIES, but finish.

vinaygaba (approved):
Thanks for contributing this!

PR #223, opened 2022-02-21
Issue 220 Fix inverse mode card when system is in dark mode

marcorighini:
Schermata 2022-02-21 alle 17 21 17


marcorighini:
Hii @vinaygaba , do you think this could be included in next release? Thanks!

vinaygaba:
@marcorighini This is definitely a really useful improvement. I'm wondering if it makes sense to make this more deterministic - instead of doing inverse, we should explicitly show the other card variants with `UI_MODE_NIGHT_NO` set.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseComponentDetailScreen.kt
```diff
@@ -3,49 +3,23 @@ package com.airbnb.android.showkase.ui
...
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.sizeIn
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.*
```
vinaygaba, line 6: Avoid wildcard imports. Might need to update you Android Studio settings since its probably happening due to default values present locally

vinaygaba: Looks like detekt also flagged this on CI, hence the build failure @marcorighini

vinaygaba (approved):
Ready to be merged once the build failures are fixed

marcorighini:
@vinaygaba fixed in d8ba947148c8f23bf6abaadd748141f21ba4e8a6

PR #226, opened 2022-03-26
Update ShowkaseBrowserActivity to handle config changes

vinaygaba:
This is to ensure that when you rotate the device, the activity isn't recreated. Its also in line with the general recommendation for Jetpack Compose. 

@airbnb/showkase-maintainers


pmecho (approved):
@vinaygaba could you share where this guidance is?
> Its also in line with the general recommendation for Jetpack Compose.

vinaygaba:
@pmecho Can't find it documentation but has come up numerous times in conversations.

Screen Shot 2022-03-29 at 12 28 05 PM

Screen Shot 2022-03-29 at 12 26 15 PM

Screen Shot 2022-03-29 at 12 26 54 PM

PR #229, opened 2022-03-28
Upgrade showkase to the latest stable releases

vinaygaba:
@airbnb/showkase-maintainers


build.gradle
```diff
@@ -108,16 +107,6 @@ allprojects {
         google()
         mavenCentral()
         jcenter()
-        maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
-    }
-    if (project.hasProperty('useKsp')) {
```
vinaygaba, line 113: Had to duplicate this in every` build.gradle` file to get around this build failure that I was seeing locally and on CI - https://github.com/airbnb/Showkase/runs/5713525186?check_suite_focus=true. Wasted too much time to fix this but eventually gave in and just used the workaround that worked for me

pmecho: Oof I do not know of a way but the workaround seems fine for now.

pmecho (approved):

Issue #207, opened 2022-01-15
Integration with Paparazzi screenshot testing


marcorighini:
Showkase is actually providing an integration with facebook screenshot testing. 
Is it possible to integrate Showkase with Paparazzi (screenshot testing without device, link https://github.com/cashapp/paparazzi)?

vinaygaba:
Showkase itself doesn't provide any integrations (yet). The blog post that you probably saw(https://proandroiddev.com/automatic-screenshot-testing-for-all-your-compose-previews-6add202fecc7) used Showkase for the automated infra but then integrated it with Facebook's library. In theory, it can work with anything but you will have to do that setup today. I fully intend to provide a sample integration and expose it as an artifact at some point.

If you are interested, I'd be happy to accept a contribution that adds a Paparazzi integration submodule (or even Shot).

davidvavra:
Wow, Paparazzi looks interesting. I'm not very happy with Facebook's library. It's not very maintained and it doesn't work for API 28+. If it runs without emulator, it can faster and a deal-breaker for CI integrations. Emulators on CI are pain to setup. I might try to play with Papparazi + Showkase in some free time.

vinaygaba:
I think that's a Compose requirement itself. I haven't double checked this but that's what I always assumed.

davidvavra:
Well, Papparazi has to do it somehow. They support snapshotting Views without instrumentation. They don't support Compose now, but they plan it for v1. But I'm assuming wrapping composables into Views might work. Might be worth investigating how they do it.

marcorighini:
I started the integration of paparazzi 1.0.0 (snapshot, the version supporting compose) in https://github.com/marcorighini/Showkase/tree/paparazzi-screenshot-testing-integration

Still to investigate:
- codegen not run on `MyScreenshotTest` file (unit test file)
- remove `ComposeTestRule` from current codegen and assess if needs to be replaced with `Paparazzi` rule

alexvanyo:
I was able to utilize Showkase's code generation to automatically populate Paparazzi tests by avoiding the `showkase-screenshot-testing` module.

Instead, I just pulled out the list of components (from `Showkase.getMetadata().componentList`), and used that as the input to a parameterized Paparazzi test, using the `componentKey` for the test name and `component` as the `@Composable` under test.

vinaygaba:
[...]

PR #232, opened 2022-04-17
Add screenshot testing integration with Shot to make introducing screenshot testing in a codebase trivial

vinaygaba:
This PR introduces a new artifact (`showkase-screenshot-testing-shot`) that integrates Showkase's automated screenshot testing artifact (`showkase-screenshot-testing`) with the [Shot](https://github.com/pedrovgs/Shot) library to provide a default fully functional implementation of automated screenshot testing for your codebase. 

In addition, the sample module also leverages this new artifact and uses it to screenshot test all the composables in the sample module. Its being tested across three different API levels (26, 28 & 29) and 2 different targets (default, google_apis), similar to the other UI tests. 

The golden copies of screenshots were also generated and uploaded as part of this PR and the artifacts (report when screenshots are being verified and the screenshots themselves when its in record mode) are now downloadable using the `upload-artifact` action. Here's what the report looks like - 

showkase-shot-demo

**Usage**

Adding screenshot testing to your codebase with Showkase + Shot requires very few steps. Assuming that Showkase was already setup in your codebase using the steps mentioned [here](https://github.com/airbnb/Showkase#installation), you will need to do the following things - 

1. Add the Showkase + Shot screenshot testing dependencies in the `build.gradle` file of the module where you defined your `@ShowkaseRoot` from the original setup

```groovy
plugins {
    ...
    id 'shot'
}

defaultConfig {
    testInstrumentationRunner "com.karumi.shot.ShotTestRunner"
}

androidTestImplementation "com.airbnb.android:showkase-screenshot-testing-shot:version"
```

2. Create an abstract class that extends `ShotShowkaseScreenshotTest` with the `ShowkaseScreenshot` annotation. Make sure that you pass the `Showkase` root class that you used when you setup Showkase originally. 

```kotlin
@ShowkaseScreenshot(rootShowkaseClass = MyTestRootModule::class)
abstract class MyScreenshotTest: ShotShowkaseScreenshotTest()
``` 

3. Use the following commands to run your screenshot tests (these are Shot commands itself). Since Shot is now configured at this point, you can use any of the commands that you see in the documentation [here](https://github.com/pedrovgs/Shot). I'm merely using the most basic example below to run the screenshot tests

```
// For recording screenshots for the very first time. You will also need to pull the screenshot that this step generated and 
// upload it in the appropriate location in your codebase so that when you run the tests, it can use these screenshots to
// compare against. Look here for an example - 
./gradlew executeScreenshotTests -Precord

// For running the screenshot tests
./gradlew executeScreenshotTests
``` 
4. That's it! All your Jetpack Compose elements that are configured to run with Showkase (including any composables that are annotated with @Preview) will automatically be screenshot tested for you! 

@airbnb/showkase-maintainers


showkase-screenshot-testing-shot/src/main/java/com.airbnb.android.showkase.screenshot.testing.shot/ShotShowkaseScreenshotTest.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    override fun onScreenshot(
+        id: String,
+        name: String,
+        group: String,
+        styleName: String?,
+        screenshotType: ShowkaseScreenshotType,
+        screenshotBitmap: Bitmap
+    ) {
+        compareScreenshot(composeTestRule)
```
badoualy, line 18: * `compareScreenshot` also takes a `name` as a parameter, we could pass a concatenation of `name/group/styleName`.
* Also has an overload taking a `Bitmap` instead of the `composeTestRule`

badoualy: Another note, I did a similar integration manually on a device running api 31, and I got the following error regarding permissions:

```
Execute ExampleShowkaseScreenshotTest_ShowkaseCodegen.composable_screenshot_test_2: FAILED
junit.framework.AssertionFailedError: junit.framework.AssertionFailedError: Failed to grant permissions, see logcat for details
```

Leaving `onScreenshot` empty still causes the error, so I guess this is caused by Showkase. Maybe worth opening an issue.
Shot seems to be handling this well, even on Android 11+ where `WRITE_EXTERNAL_STORAGE` is deprecated.

If anyone has the same issue, in the mean time can be fixed by overloading `takeComposableScreenshot` functions to bypass showkase screenshot process:

```kotlin
@ShowkaseScreenshot(rootShowkaseClass = RootModule::class)
open class ExampleShowkaseScreenshotTest : ShowkaseScreenshotTest, ScreenshotTest {
    @get:Rule
    override val composeTestRule = createComposeRule()

    // Do the same for typo/color functions
    override fun takeComposableScreenshot(showkaseBrowserComponent: ShowkaseBrowserComponent) {
        // Disable animations for screenshots to make them deterministic
        composeTestRule.mainClock.autoAdvance = false
        composeTestRule.setContent { showkaseBrowserComponent.component() }
        compareScreenshot(
            composeTestRule,
            name = showkaseBrowserComponent.run { "${group}_${componentName}_${styleName}" }
        )
    }

    override fun onScreenshot(
        id: String,
        name: String,
        group: String,
        styleName: String?,
        screenshotType: ShowkaseScreenshotType,
        screenshotBitmap: Bitmap
    ) {
        // no-op
    }
}
```

badoualy: The `ShowkaseScreenshotTest` changes in this branch https://github.com/marcorighini/Showkase/commit/ee3176577b357cdbacb832ab4e6a2e93e3a5f6f4#diff-eefe4c579476fb035be5abc91fcb8701101f66f9b9cd494e774f4f89d83ccf20 would be simplify this, and we wouldn't need to override the 3 functions to bypass current screenshot mechanism

pmecho: Maybe @marcorighini can contribute those changes back into this and we could also create a `showkase-screenshot-testing-paparazzi` module? Thoughts @vinaygaba?

pmecho (approved):

showkase-screenshot-testing-shot/src/main/java/com.airbnb.android.showkase.screenshot.testing.shot/ShotShowkaseScreenshotTest.kt
vinaygaba: @badoualy for the error that you were seeing, I added this in the Manifest and that seems to solve the problem when I tested it - https://github.com/airbnb/Showkase/pull/232/files#diff-c53f7bc57c4a7ffb92824cc2598ff253042b28a0d38f69eead012e62f7c41936R9

@pmecho As for Paparazzi, my follow up PR is going to be the `showkase-paparazzi` artifact as I'm very excited about it. It will dramatically simplify things as we can simply rely on parametrized testing instead of leveraging code generation. The reason I had to use code generation in the first place was that Firebase Test Labs did not work well with parameterized tests due to something to do with sharding iirc. 

The pararazzi implementation will look very similar to what Alex describes in this comment - https://github.com/airbnb/Showkase/issues/207#issuecomment-1114042072

badoualy: @vinaygaba `requestLegacyExternalStorage` no longer works when targeting api 30+
https://developer.android.com/training/data-storage/use-cases#opt-out-in-production-app

PR #234, opened 2022-05-02
Fix illegal chars getting generated into function names.

rossbacher:
Illegal chars might get generated into function names which compile in Kotlin but are not allowed in dex and will lead to build errors.

### DIFF
--- a/sample/src/main/java/com/airbnb/android/showkasesample/Text.kt
+++ b/sample/src/main/java/com/airbnb/android/showkasesample/Text.kt
@@ -143,3 +143,14 @@ fun H6TextRowComponentPreview(
 ) {
     H6TextRowComponent(person.name)
 }
+
+// This is here for testing, generating this code into function names would
+// fail dex generation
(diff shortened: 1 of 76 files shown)


rossbacher:
Will do that in a bit!

> LGTM! You will need to update the tests as well. You can toggle this boolean and then run it locally to update the test sources - https://github.com/airbnb/Showkase/blob/master/showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/BaseProcessorTest.kt#L18

Will that that in a bit, thx!

rossbacher:
> Will do that in a bit!
> 
> > LGTM! You will need to update the tests as well. You can toggle this boolean and then run it locally to update the test sources - https://github.com/airbnb/Showkase/blob/master/showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/BaseProcessorTest.kt#L18
> 
> Will that that in a bit, thx!

@vinaygaba done, can you have one more look if that is ok that way?

vinaygaba (approved):
@rossbacher Since this impacts only the generated code, this is an acceptable way to name the functions!

PR #238, opened 2022-06-24
Specify the type names of the generated ShowkaseProvider functions (explicit-api support)

PaulWoitaschek:
In order to support kotlins explicit api, the return types must be specified. Additionally enables explicit-api in the processor tests to catch future regressions.


vinaygaba (approved):
Very cool. Thank you for your contribution 🙏

PR #240, opened 2022-07-07
Fix textoverflow on top appbar for components with long names

oas004:
Thank you for a very cool library! I absolutely love this! 😄 😍 When I started using it I noticed that the names of our components where quite long and that the toolbar was not quite handling this as you can see from the screenshots. 

I had some trouble testing this and would love some help and feedback on that. I tried to make an Android test in the `:showcase-browser-testsing` but could not quite figure out a good way to test it. I also thought that it might be a bit flaky since the overflow will change with regards to screen size. 

### Before
toolbar_large_title

### Now
toolbar_large_title_truncated

### DIFF
--- a/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
+++ b/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
@@ -31,6 +31,7 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import androidx.navigation.NavHostController
@@ -91,8 +92,7 @@ internal fun ShowkaseAppBar(
     Row(
         Modifier.fillMaxWidth()
             .graphicsLayer(shadowElevation = 4f)
-            .padding(padding2x)
-            .height(64.dp),
+            .padding(padding2x),
         horizontalArrangement = Arrangement.SpaceBetween,
         verticalAlignment = Alignment.CenterVertically
     ) {
@@ -198,7 +198,9 @@ fun ToolbarTitle(
             fontSize = 20.sp,
             fontFamily = FontFamily.Monospace,
             fontWeight = FontWeight.Bold
-        )
+        ),
+        maxLines = 3,
+        overflow = TextOverflow.Ellipsis
     )
 }
 
--- a/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
+++ b/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
@@ -7,7 +7,6 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.Icon
 import androidx.compose.material.IconButton
@@ -21,6 +20,8 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -28,11 +29,13 @@ import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import androidx.compose.ui.semantics.SemanticsPropertyReceiver
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextOverflow
-import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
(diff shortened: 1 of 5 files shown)


vinaygaba:
@oas004 Could you upload the screenshots again. I'm not able to see them

Screen Shot 2022-07-07 at 9 11 30 PM

vinaygaba:
Re: Testing

You can add a composable [here](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/main/java/com/airbnb/android/showkase_browser_testing/TestComposables.kt) that has a really long value for `name` in the `ShowkaseComposable` or `Preview` annotation (a name that extends at least 2 lines). Then you can add a [test case in this file](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt) that verifies that the entire name is visible. You are right about the different screen sizes generally speaking but the testing we are doing in this repo are restricted to a single device as defined in the [android.yml](https://github.com/airbnb/Showkase/blob/master/.github/workflows/android.yml) file so it should be fine to add a test case.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -91,8 +92,7 @@ internal fun ShowkaseAppBar(
     Row(
         Modifier.fillMaxWidth()
             .graphicsLayer(shadowElevation = 4f)
-            .padding(padding2x)
-            .height(64.dp),
+            .padding(padding2x),
```
vinaygaba, line 95: we might want to add a maxSize for the height but I'm curious to see the screenshots before that

oas004:
> @oas004 Could you upload the screenshots again. I'm not able to see them
> 
> Screen Shot 2022-07-07 at 9 11 30 PM

Hi, thanks for quick review! 😄 I tried to replace the screenshots now and upload them manually(instead of dragging and dropping them). Does it work now? :)

oas004:
> Re: Testing
> 
> You can add a composable [here](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/main/java/com/airbnb/android/showkase_browser_testing/TestComposables.kt) that has a really long value for `name` in the `ShowkaseComposable` or `Preview` annotation (a name that extends at least 2 lines). Then you can add a [test case in this file](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt) that verifies that the entire name is visible. You are right about the different screen sizes generally speaking but the testing we are doing in this repo are restricted to a single device as defined in the [android.yml](https://github.com/airbnb/Showkase/blob/master/.github/workflows/android.yml) file so it should be fine to add a test case.

Thanks for quick review! Actually, when trying to write a test case for this I ran in to an issue. If I try to make a test on the node with test tag that the text is ellipsis, the NodeText has the entire text. So that means the text will go green if the text is correctly clipped an if It is not. There is an issue on this: https://issuetracker.google.com/issues/238332005 :) Could you please provide me with guidance on how to write a test case for this? :)

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -186,19 +189,34 @@ private fun ShowkaseAppBarTitle(
     }
 }
 
+val LineCountKey = SemanticsPropertyKey<Int>("lineCount")
```
oas004, line 192: Should this be in its own file? :) What do u think?

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
oas004: hmm yeah maybe you're right 🤔 Is it also maybe a bit weird that it spans 3 lines instead of 2? What do you think? :)

oas004:
> > Re: Testing
> > You can add a composable [here](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/main/java/com/airbnb/android/showkase_browser_testing/TestComposables.kt) that has a really long value for `name` in the `ShowkaseComposable` or `Preview` annotation (a name that extends at least 2 lines). Then you can add a [test case in this file](https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt) that verifies that the entire name is visible. You are right about the different screen sizes generally speaking but the testing we are doing in this repo are restricted to a single device as defined in the [android.yml](https://github.com/airbnb/Showkase/blob/master/.github/workflows/android.yml) file so it should be fine to add a test case.
> 
> Thanks for quick review! Actually, when trying to write a test case for this I ran in to an issue. If I try to make a test on the node with test tag that the text is ellipsis, the NodeText has the entire text. So that means the text will go green if the text is correctly clipped an if It is not. There is an issue on this: https://issuetracker.google.com/issues/238332005 :) Could you please provide me with guidance on how to write a test case for this? :)

Managed to find a workaround for the test. Added it in [baea74a](https://github.com/airbnb/Showkase/pull/240/commits/baea74a682aa9bd1129f80d29b9e06e9ae12ac7c)  😄 What do you think?

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -186,19 +189,34 @@ private fun ShowkaseAppBarTitle(
...
+var SemanticsPropertyReceiver.lineCountVal by LineCountKey
+
 @Composable
 fun ToolbarTitle(
     string: String,
     modifier: Modifier
 ) {
+    val lineCount = remember {
```
vinaygaba, line 200: Not ideal to have test logic live in the component directly but I guess that happens pretty often in Compose due to how semantics are setup with some hooks into testing. Don't have a better idea on testing this so might be fine for now.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
vinaygaba: Maybe move to a file called `SemanticsUtils` or something along those lines

vinaygaba (approved):
Great work. One minor request and we are good to merge this after that.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
oas004: Good idea! Moved it into `SemanticsUtils` in [833e234](https://github.com/airbnb/Showkase/pull/240/commits/833e234c90e82b0070ac40dbab265f4312918787) 😄

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
oas004: Totally agree with you on that! I think this can work for now :)

PR #241, opened 2022-07-13
Format codegen from long named preview parameter provider so it compiles

### LINKED ISSUE #192: Long lines in generated code is formatted incorrectly, leading to compilation error
I'm getting compilation errors in `ShowkaseCodegen`. Lambda start brackets seem to be placed on the next line, which doesn't compile:
```kotlin
my.very.long.package.name.ViewStateProvider().values.iterator().asSequence().forEachIndexed
    { index, previewParam -> 
        // ...
}
```

oas004:
Thanks for an awesome library! I really like this! Good job! 😃 🤩 

I found that if you have a long named preview parameter provider, the code that was generated was not formatted correctly and resulting in a compile error. Before, the lambda would start the bracket on the next line and it would not compile. Now, I have the lambda start on the next line and formatted the line. I think this should fix #192 also :)

Before
```kotlin
MyProvider().values.iterator.asSequence.forEachIndexed
{ index, previewParam ->
```

Now
```kotlin
MyProvider()
    .values
    .iterator
    .asSequence.forEachIndexed { index, previewParam ->
```

I have added a test for this as well in the `:showkase-processor-testing` module. Is that enough?


oas004:
@vinaygaba Rebased and fixed tests according to #238 😄

vinaygaba (approved):
Fantastic improvement. Apologize for the delay in reviewing it. Really grateful for all the contributions you've made lately ❤️

PR #243, opened 2022-07-16
Remove redundant code in `ShowkaseTheme.kt`

oas004:
Thanks for an awesome library! I noticed that there was some redundant code in the `ShowkaseTheme.kt` code file, so I optimised it In this PR. 😄

### DIFF
--- a/sample/src/main/java/com/airbnb/android/showkasesample/ShowkaseTheme.kt
+++ b/sample/src/main/java/com/airbnb/android/showkasesample/ShowkaseTheme.kt
@@ -5,10 +5,9 @@ import androidx.compose.material.MaterialTheme
 import androidx.compose.material.darkColors
 import androidx.compose.material.lightColors
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.res.colorResource
 
 @Composable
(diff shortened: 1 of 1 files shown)


sample/src/main/java/com/airbnb/android/showkasesample/ShowkaseTheme.kt
```diff
@@ -5,10 +5,9 @@ import androidx.compose.material.MaterialTheme
 import androidx.compose.material.darkColors
 import androidx.compose.material.lightColors
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.res.colorResource
 
 @Composable
-fun ShowkaseTheme(children: @Composable()() -> Unit) {
```
vinaygaba, line 11: ha this rings a bell. Showkase was started when Compose was still in dev preview. At the time, this sytax used to be needed 😅

vinaygaba (approved):

sample/src/main/java/com/airbnb/android/showkasesample/ShowkaseTheme.kt
oas004: Aha! That makes sense! 😄

PR #246, opened 2022-07-26
Add Icons page and support `@ShowkaseIcon`

### LINKED ISSUE #245: [IDEA]  Showkase Icons page on the landing page
Thanks for an awesome library! 😄 I really like it! 🤩 

When using it in our component library I thought that it would be cool to display all our icons somehow. I think we could wrap them all in a composable function and add the `@ShowkaseComposable` attribute and it will display inside the components page. However, would it be cool if we had a page only for the icons like we have for typography and colors? 😃 Any thoughts? 

Feel free to close this if you think it should be part of the components page 😃

oas004:
### Feature description

First of all, thanks for an amazing cool library! I really like it! 🤩 When using it I noticed that it would be nice to have like a gallery of our icons as well in the same context, so I added a suggestion in this PR on how it could look. 

In this PR I have added support for displaying `ImageVectors` and `@DrawableRes Int`as Icons in its own page.

It can be used like this inside a class or an object:

```kotlin
@ShowkaseIcon(name = "Name", group = "Group")
val searchIcon = Icon.Filled.SearchIcon
```

or 

```kotlin
@ShowkaseIcon(name = "Name", group = "Group")
@DrawableRes
val searchIcon = R.drawable.search_icon
```

This fixes #245 😄 

### Things to consider

I have added tests in the `showkase-processor-testing` and `showkase-browser-testing`. Are there any more tests that needs to be added?

I was thinking about adding an annotation parameter for the tint. Something like

```kotlin
@ShowkaseIcon(name = "Name", group = "Group", tint = Color)
val searchIcon = Icon.Filled.SearchIcon
```

But I'm not quite sure if that will work and if the best way here would be to provide a HEX code or something like that? Do you have any input on this?

### Gameplay 🎮 

Here is some gameplay from the sample app

arrows
landing_icons
landing
launcher
verification


oas004:
I have some issues with generating the screenshots for the screenshot testing of the Icons 🤔 I can't find a way to check the type of the icon in AndroidTestDebug. Could you help me out a bit @vinaygaba ? 😄

oas004:
> I have some issues with generating the screenshots for the screenshot testing of the Icons 🤔 I can't find a way to check the type of the icon in AndroidTestDebug. Could you help me out a bit @vinaygaba ? 😄

I think I got this to work now 😄 🥳 A bit unsure if this is the best solution, but would like some feedback on this if someone has time 😄

oas004:
Rebased and updated the test from #241 😄 Looks like the UI tests are being cancelled. Do you know if there is a reason for it being cancelled @vinaygaba ? 😄

oas004:
Im thinking, would it be nice to be able to annotate the `object` and get all the icon that it encapsulates? 🤔 I think the api as it is now, might be a bit difficult to use as companies might have 100 of different icons 🤔 😄 Wdyt?

oas004:
@vinaygaba I am thinking about closing this as it is hard to maintain the branch with all the merge conflicts. Is that okay? Then we can maybe discuss a better way to make the API for this when the multi preview stuff is done? :) Can keep the branch for reference maybe? :)

vinaygaba:
@oas004 Yes I agree! would be nice to brainstorm a bit so that we can not only solve it for icons, but also bring improvements to colors and typography.

PR #253, opened 2022-08-18
Bump compose compiler to 1.3.1 and libs to 1.2.1

oas004:
In this PR I have bumped the Compose compiler version to 1.3.0 and the libs to 1.2.1 as the latest stable version 😄

This also meant I had to bump the following:

- KSP -> 1.7.10-1.0.6
- Compose activity -> 1.5.1
- Kotlin -> 1.7.10
- KotlinCompilerVersion -> 1.7.10

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,13 +1,14 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
-    ext.KSP_VERSION = '1.7.0-1.0.6'
+    ext.KSP_VERSION = '1.7.10-1.0.6'
 
     ext.versions = [
             'androidXTestCore'       : '1.4.0',
             'androidXTestRules'      : '1.4.0',
             'assertJ'                : '3.16.1',
-            'compose'                : '1.2.0',
-            'composeActivity'        : '1.5.0',
+            'compose'                : '1.2.1',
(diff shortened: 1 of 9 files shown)


oas004:
I guess 1.3.1 is out now? 🤔 @vinaygaba Should I close this or should I try to update to 1.3.1? Wdyt? :)

vinaygaba:
Maybe do it in this PR itself @oas004. Since they both support the same version of kotlin, it should be fine to use the latest one. I hadn't merged this before because I wanted to do a release first (which we already did last month) to make sure people using the older version of Kotlin had a version of Showkase that would work. I think going forward, it's compatible regardless so should be easier for us to do this version bumps

oas004:
> Maybe do it in this PR itself @oas004. Since they both support the same version of kotlin, it should be fine to use the latest one. I hadn't merged this before because I wanted to do a release first (which we already did last month) to make sure people using the older version of Kotlin had a version of Showkase that would work. I think going forward, it's compatible regardless so should be easier for us to do this version bumps

Aha yes that makes sense! I will try to update this PR then 😄

oas004:
@vinaygaba Bumped to target 1.3.1 now :) Tests seems fine and seems fine locally :)

vinaygaba (approved):

Issue #233, opened 2022-04-30
Support multi-preview annotations (annotations annotated with @Preview)


alexvanyo:
In the latest alpha releases of Compose, the `@Preview` annotation can now be applied to other annotation classes: https://android-review.googlesource.com/c/platform/frameworks/support/+/1964936. This will cause those other annotation classes to be treated as previews. This is most useful for bundling common sets of previews together (since the custom annotation class can be annotated with multiple `@Preview`s), instead of having to copy around multiple groups `@Preview` everywhere.

Right now, the Showkase process throws an exception of "Only composable methods can be annotated with Preview" upon seeing a custom annotation annotated with `@Preview`.

vinaygaba:
Thanks for flagging this! I think I should be able to skip this check when its applied to an annotation class. I believe I should have this information available and should be a relatively straightforward fix!

oas004:
When trying to add the last step for supporting multi-preview annotations I stumbled on two issues that I would love to have a discussion on.  Let's say you have a custom annotation like:

```
@Preview(name = "phone", device = "spec:shape=Normal,width=360,height=640,unit=dp,dpi=480")
@Preview(name = "foldable", device = "spec:shape=Normal,width=673,height=841,unit=dp,dpi=480")
@Preview(name = "tablet", device = "spec:shape=Normal,width=1280,height=800,unit=dp,dpi=480")
annotation class CustomDevicePreview
```

For this type of custom annotation, IMO it will not make sense to render 4 showkase previews for each time you use it on a composable. That is because they would all look the same since Showkase is running on an actual device. However, let's say you have something like this

```
@Preview(name="Composable in 400dp width", widthDp=400)
@Preview(name="Composable in 200dp width", widthDp=200)
annotation class CustomSizePreview
```

IMO, in this case it would make sense for showkase to render both permutations for each usage of the annotation. Here the showkase previews will actually differ in contrast to the first case.

Do you think we should provide a way for the developer to choose if they would like to render all permutations or only one? Wdyt? :) 

The second issue I ran into was in regards to the naming of these components. Since we are using the preview names, we risk getting a lot of components with the same name. I tried fixing this by adding the element name as well. However, we would still get the same component names if the user do not provide any preview names.  If I add the index to the name it looks very ugly. Do you have any input on how to solve this more elegantly?  🤔 😄 Any input is appreciated 😄

alexvanyo:
[...]

PR #255, opened 2022-08-22
Make multi-preview annotations not throw at compiletime

oas004:
This is the first PR in order for Showkase to support multi-preview annotations (annotations annotated with `@Preview`)(#233)

In this PR I have only made sure that consumers that are using multi-preview annotations can also use Showkase without compiletime failures. If the annotation processor faces such an issue, it will just skip it for now.

Leaving this as draft, as I have some questions about the validator function 😅

### DIFF
--- a/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
+++ b/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
@@ -30,6 +30,7 @@ public annotation class DevicePreviews
 public class Composables {
 
     @FontScalePreviews
+    @Preview(name = "component", group = "component-group")
     @Composable
     public fun Component() {
     }
--- a/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
+++ b/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
@@ -1,0 +1,25 @@
+package com.airbnb.android.showkase_processor_testing
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+
+
+@Preview(
+    name = "small font",
+    group = "font scales",
+    fontScale = 0.5f
+)
+@Preview(
+    name = "large font",
+    group = "font scales",
+    fontScale = 1.5f
+)
+public annotation class FontScalePreviews
+
+public class Composables {
+
+    @FontScalePreviews
+    @Composable
+    public fun Component() {
+    }
+}
--- a/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
+++ b/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
@@ -3,7 +3,6 @@ package com.airbnb.android.showkase_processor_testing
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.tooling.preview.Preview
 
-
 @Preview(
     name = "small font",
     group = "font scales",
--- a/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
+++ b/showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
@@ -1,6 +1,8 @@
 package com.airbnb.android.showkase_processor_testing
 
+import android.content.res.Configuration.UI_MODE_NIGHT_YES
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.semantics.SemanticsProperties.Text
 import androidx.compose.ui.tooling.preview.Preview
 
 @Preview(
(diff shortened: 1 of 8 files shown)


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -102,10 +103,14 @@ class ShowkaseProcessor @JvmOverloads constructor(
                     element,
                     PREVIEW_SIMPLE_NAME
                 )
-                getShowkaseMetadataFromPreview(
-                    element = element,
-                    showkaseValidator = showkaseValidator
-                )
+                if (element.isMethod()) {
```
oas004, line 106: Im having some doubts about this here. Can I do this? 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
```diff
@@ -75,6 +77,13 @@ internal class ShowkaseValidator {
         }
     }
 
+    // This should check if it is an annotation with multiple @Preview annotations
+    private fun checkElementIsMultiPreview(element: XElement): Boolean {
+       return element.getAllAnnotations().size >= 2 && element.findAnnotationBySimpleName(
+                PREVIEW_SIMPLE_NAME) != null && element.findAnnotationBySimpleName(
+           METADATA_SIMPLE_NAME) == null
+    }
```
oas004, line 85: I'm very unsure about this method. For me it made sense as in this step for the multi.preview annotation, it has more than or two annotations, at least there should be a preview and the metadata annotation should not be present yet 🤔 However, Im having some doubts if it is correct. Is there a better way to do this?

vinaygaba: Why are we looking for the Metadata annotation? Does that denote that it's an annotation?

vinaygaba: Also, we don't necessarily need the size >= 2 check. While that's probably going to be true in most cases, we don't need to prescribe that since it's not a requirement for multi-preview. You can annotate another annotation with a single `@Preview` so no reason to add an extra requirement.

vinaygaba:
cc @alexvanyo I'm working with Odin in getting Multi-preview support to Showkase.

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
oas004: That is a good point! Maybe I can check its the element is an annotation class type? 🤔

vinaygaba: Yeah that's probably a better direction to go in

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
```diff
@@ -75,6 +75,11 @@ internal class ShowkaseValidator {
         }
     }
 
+    // This should check if it is an annotation with multiple @Preview annotations
```
vinaygaba, line 78:
```suggestion
    // This should check if it is an annotation that's annotated with @Preview annotation
```

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
```diff
@@ -39,20 +39,20 @@ internal class ShowkaseValidator {
         }
 
         when {
-            !element.isMethod() -> {
+            !element.isMethod() && !checkElementIsMultiPreview(element) -> {
```
vinaygaba, line 42: Since we return early [here](https://github.com/airbnb/Showkase/pull/255/files#diff-d398a33999c90a9159642dd79319c9723282054253bbbf95b5b75b89d9d5cd9eR106), we shouldn't need any updates in this file no?

oas004: Might be me that is missing something here, but wouldn't we need to to this check here anyways since the validator is run before that point? 🤔 😃

oas004: Aha! Thought about this, and it makes sense to do an early return as you said 😄 Didn't see that at first. Changed it up in c57de47  and reverted the validator changes 😄 That was quite clever! Thanks for that!

showkase-processor-testing/src/test/resources/ShowkaseProcessorTest/composable_function_with_multiple_preview_functions_compiles/input/Composables.kt
```diff
@@ -0,0 +1,24 @@
+package com.airbnb.android.showkase_processor_testing
```
vinaygaba, line 1: You will need to generate the output of this function as well. You can do that by making this boolean true and running the test locally - https://github.com/airbnb/Showkase/blob/master/showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/BaseProcessorTest.kt#L18

This should generate the relevant output files that you can then commit to this PR

oas004: I might have misunderstood something here, but my understanding was that this first PR just needed to get it not to throw at compile time. So that is why Im not making a ShowkaseMetadata object when processing the annotation. Then I don't think the test will generate an output file. My question is then, should I make this into multiple PRs or should I try to achieve full on support in this one? 😄 wdyt? 

Furthermore, the test flag was a really nice trick! 😄I totally missed that. Would it be nice to add a comment on that in the contribution guide? wdyt? 😄

vinaygaba: @oas004 Oh I see. It won't generate an output because your input file does not have any @Preview annotation on a method. Maybe just add one to the same file to ensure that it skips multi preview but correctly picks up previews when they are annotated to a function.

oas004: Oo yeah good idea! Added one in [7ac8862](https://github.com/airbnb/Showkase/pull/255/commits/7ac8862f35526767a4a857aa47bd7c87ebf9e98d) 😄 Does that look okay? 😄

vinaygaba (approved):

PR #259, opened 2022-08-28
Make stacked preview annotations generate showkaseMetadata components for each annotation

oas004:
This is the next step in supporting Multipreview annotations. Referencing #255 as the first step. 

Here I have added functionality to make a `ShowkaseMetadata` object for each `Preview` annotation. Eg.

```
@Preview(
    name = "small font",
    group = "font scales",
    fontScale = 0.5f
)
@Preview(
    name = "large font",
    group = "font scales",
    fontScale = 1.5f
)
@Composable
fun ComposablePreviewFont() {

}

```

Before, this would only make one component in Showkase. Now it should make one for each preview and place them in their respective group. Here they would both be placed in the group `font scales`

Furthermore, I would advice to review this by commit as I had to update the tests because of some arrangement due to a check in the processor. I placed the test update in 8371ec2efd81e6317c85029e377dca3ae8cd0aee.

I'm leaving this as a draft PR because I'm having some issues with getting this to work with KAPT. 
With KSP, the method `roundEnvironment.getElementsAnnotatedWith(PREVIEW_CLASS_NAME)` returns all the elements for multi stacked preview. However, with KAPT, it does not seem to work the same way. Is there some other function I should use here? Any suggestions are more than welcome 😄 
Added a [slack post](https://kotlinlang.slack.com/archives/C2R77UD35/p1661703541100529) for anyone interested 😄

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -56,7 +56,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -DuseKsp=false
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=false -DuseKsp=false
 
       - name: Run UI Tests w/ KSP
         uses: reactivecircus/android-emulator-runner@v2
@@ -65,7 +65,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -DuseKsp=true
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=true -DuseKsp=true
 
       - name: Run Screenshot Tests
         uses: reactivecircus/android-emulator-runner@v2
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -56,7 +56,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=false -DuseKsp=false
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=false
 
       - name: Run UI Tests w/ KSP
         uses: reactivecircus/android-emulator-runner@v2
@@ -65,7 +65,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=true -DuseKsp=true
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=true
 
       - name: Run Screenshot Tests
         uses: reactivecircus/android-emulator-runner@v2
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -56,7 +56,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=false
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -DuseKsp=false
 
       - name: Run UI Tests w/ KSP
         uses: reactivecircus/android-emulator-runner@v2
@@ -65,7 +65,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=true
+          script: ./gradlew connectedCheck --no-daemon --stacktrace -DuseKsp=true
 
       - name: Run Screenshot Tests
         uses: reactivecircus/android-emulator-runner@v2
--- a/sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
+++ b/sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
@@ -57,3 +57,14 @@ fun BasicChipBluePreview() {
         modifier = Modifier.background(color = Color.Blue)
     )
 }
+
+
+@ShowkaseComposable(name = "Basic Chip Grey Light", group = "Chips")
+@ShowkaseComposable(name = "Basic Chip Grey Dark", group = "Chips")
+@Composable
+fun BasicChipGreyPreview() {
+    BasicChip(
+        text = "Chip Component",
+        modifier = Modifier.background(color = Color.Gray)
+    )
+}
--- a/sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
+++ b/sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
@@ -1,5 +1,7 @@
 package com.airbnb.android.showkasesample
 
+import android.content.res.Configuration.UI_MODE_NIGHT_NO
+import android.content.res.Configuration.UI_MODE_NIGHT_YES
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.padding
@@ -9,6 +11,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.airbnb.android.showkase.annotation.ShowkaseComposable
 
@@ -42,5 +45,15 @@ fun BasicChipYellowPreview() {
     BasicChip(
         text = "Chip Component",
         modifier = Modifier.background(color = Color.Yellow)
+    )
+}
+
+@Preview(name = "Basic Chip Blue Light", group = "Chips", uiMode = UI_MODE_NIGHT_NO)
+@Preview(name = "Basic Chip Blue Dark", group = "Chips", uiMode = UI_MODE_NIGHT_YES)
+@Composable
+fun BasicChipBluePreview() {
+    BasicChip(
+        text = "Chip Component",
+        modifier = Modifier.background(color = Color.Blue)
     )
 }
--- a/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
+++ b/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
@@ -47,6 +47,11 @@ class ShowcaseBrowserTest {
         7
     }
 
+    @Before
+    fun setupLog() {
+        println("Use KSP value = ${System.getProperty("useKsp")}")
+    }
+
     @Test
     fun activity_starts_and_all_the_showkase_ui_elements_are_visible_on_the_screen_and_clickable() {
         // Assert that all the categories are displayed on the screen and that they are clickable.
--- a/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
+++ b/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
@@ -6,7 +6,6 @@ import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.platform.app.InstrumentationRegistry
 import com.airbnb.android.showkase.models.Showkase
 import com.airbnb.android.showkase.ui.ShowkaseBrowserActivity
-import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -41,15 +40,10 @@ class ShowcaseBrowserTest {
     // beyond the source retention KEEP, except for in the new IR backend which was introduced
     // in Kotlin 1.6. It will be available in the old backend in Kotlin version 1.7.20.
     // See https://youtrack.jetbrains.com/issue/KT-49682 for more information about this.
-    private val componentSize = if (System.getProperty("useKsp") == "true") {
+    private val componentSize = if (BuildConfig.IS_RUNNING_KSP) {
         11
     } else {
         7
-    }
-
-    @Before
-    fun setupLog() {
-        println("Use KSP value = ${System.getProperty("useKsp")}")
     }
 
     @Test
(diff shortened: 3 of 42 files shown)


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -96,6 +96,7 @@ class ShowkaseProcessor @JvmOverloads constructor(
 
 
     private fun processPreviewAnnotation(roundEnvironment: XRoundEnv): Set<ShowkaseMetadata.Component> {
```
vinaygaba, line 98: I wonder if we should enable doing this repeatable annotation support even for `ShowkaseComposable` 🤔 What do you think? I'm on the fence about it.

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -52,6 +52,8 @@ internal sealed class ShowkaseMetadata {
         override val enclosingClassName: ClassName? = null,
         override val insideWrapperClass: Boolean = false,
         override val insideObject: Boolean = false,
+        val componentIndex: Int? = null,
+        val componentDistinctName: String? = null,
```
vinaygaba, line 56: since this can be inferred (`"${element.name}_$index"`), maybe we don't need to necessarily save this 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseCodegenMetadataWriter.kt
```diff
@@ -63,16 +68,22 @@ internal class ShowkaseCodegenMetadataWriter(private val environment: XProcessin
...
+        val  builder = AnnotationSpec.builder(ShowkaseCodegenMetadata::class)
             .addMember("showkaseGroup = %S", showkaseMetadata.showkaseGroup)
             .addMember("packageName = %S", showkaseMetadata.packageName)
             .addMember("packageSimpleName = %S", showkaseMetadata.packageSimpleName)
             .addMember("showkaseElementName = %S", showkaseMetadata.elementName)
             .addMember("insideObject = ${showkaseMetadata.insideObject}")
             .addMember("insideWrapperClass = ${showkaseMetadata.insideWrapperClass}")
             .addMember("showkaseKDoc = %S", showkaseMetadata.showkaseKDoc)
+        if (showkaseMetadata is ShowkaseMetadata.Component && showkaseMetadata.componentDistinctName != null) {
+            builder.addMember("showkaseName = %S", showkaseMetadata.componentDistinctName)
```
vinaygaba, line 81: won't this cause the name of the component to have index in the ShowkaseBrowser 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
oas004: Yeah that is a good point! I can just use the `ComponentIndex` instead 😄 I'll fix this, thanks!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseCodegenMetadataWriter.kt
oas004: Yeah, this I'm somewhat unsure about. In the ShowkaseBrowser, I think it makes sense to distinguish them somehow. Here, I just used the `componentDistinctName`, but maybe it makes sense to use the preview name instead? 🤔 There is no check for the consumer if they would name the preview with the same names, but I think that would maybe be the best way to distinguish them? 🤔 Wdyt?

oas004: Reverted this in [f185cec](https://github.com/airbnb/Showkase/pull/259/commits/f185cec407e6bc8103e3a378726d5b0eb01f215c) As there was no reason to have the index in the name in the ShowkaseBrowser 😄  Thanks for pointing this out!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
oas004: This should be fixed in [f185cec](https://github.com/airbnb/Showkase/pull/259/commits/f185cec407e6bc8103e3a378726d5b0eb01f215c) 😃 Thanks!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
vinaygaba: @oas004 From my perspective, the goal for `ShowkaseComposable` was to offer everything `Preview` offers + more functionality that we'd need on top of that. When I use this lens, I feel like allowing ShowkaseComposable to behave in the same manner might be valuable

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseCodegenMetadataWriter.kt
```diff
@@ -31,7 +31,16 @@ internal class ShowkaseCodegenMetadataWriter(private val environment: XProcessin
...
 
         showkaseMetadataSet.forEach { showkaseMetadata ->
-            val name = "${showkaseMetadata.showkaseGroup}_${showkaseMetadata.showkaseName}"
+
+            val name = if (
+                showkaseMetadata is ShowkaseMetadata.Component
+                && showkaseMetadata.componentIndex != null
+                && showkaseMetadata.componentIndex > 0
+            ) {
+                "${showkaseMetadata.showkaseGroup}_${showkaseMetadata.showkaseName}_${showkaseMetadata.componentIndex}"
```
vinaygaba, line 40: I know we were hoping for a better name but we can probably append `componentIndex` for now.

vinaygaba (approved):
This is looking in good shape. You can do the `ShowkaseComposable` support later on as it's not super pressing. I think you are at a point where you can move to the custom annotation stuff we discussed previously. Great job in making such a critical feature happen 👏🏻

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Yeah when seeing it from that perspective I agree. Also I think the consumer might expect the same behaviour as the `Preview` annotation 🤔 I tried to add support for it in b109238 😄 It only works with KSP as the stacked preview as this has the same issue with KAPT.

showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/ShowkaseProcessorTest.kt
```diff
@@ -508,5 +508,17 @@ class ShowkaseProcessorTest : BaseProcessorTest() {
     fun `composable function with multiple preview functions compiles`() {
         compileInputsAndVerifyOutputs()
     }
+
+    @Test
+    fun `composable function with multiple preview annotations stacked generates output`() {
+        // This functionality is only supported with KSP for now.
+        compileInputsAndVerifyOutputs(modes = listOf(Mode.KSP))
```
vinaygaba, line 515: 👍

oas004:
> This is looking in good shape. You can do the `ShowkaseComposable` support later on as it's not super pressing. I think you are at a point where you can move to the custom annotation stuff we discussed previously. Great job in making such a critical feature happen 👏🏻

Thank you very much @vinaygaba! And thank you so much valuable review 🙏 I tried to add support for stacked `@ShowkaseComposable` in b109238. I also wrote some docs about these not working with KAPT because of https://youtrack.jetbrains.com/issue/KT-49682 . Furthermore, added tests for only KSP for stacked `@Preview` and `@ShowkaseComposable` cases 😃

vinaygaba:
@oas004 Amazing! One final thing you can do is write a test for the ShowkaseBrowser that verifies the following:
- Stacked previews show up in the browser
- They are being treated as separate previews

Add the test cases here - https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt

oas004:
> @oas004 Amazing! One final thing you can do is write a test for the ShowkaseBrowser that verifies the following:
> 
> * Stacked previews show up in the browser
> * They are being treated as separate previews
> 
> Add the test cases here - https://github.com/airbnb/Showkase/blob/master/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt

That is a very good idea! 🙌  Is there a way for us to bypass it in KAPT on CI? I think if I added those BrowserTests, then it will fail on KAPT every time 🤔

vinaygaba:
@oas004 Oh no, you are right. We currently run these tests for both ksp and kapt and pass the `PuseKsp` flag from CLI. Is there a way to access this value in the test case somehow? If there is, we can always pass the test when it's `kapt` and run the test as usual when its `ksp`. This is how we run these tests - https://github.com/airbnb/Showkase/blob/master/.github/workflows/android.yml#L59

vinaygaba:
@oas004 Can you try this - in the android.yml file, also add `-DuseKsp=true` and `-DuseKsp=false` and then try accessing them in the test using `System.getProperty("useKsp")`

oas004:
> @oas004 Can you try this - in the android.yml file, also add `-DuseKsp=true` and `-DuseKsp=false` and then try accessing them in the new test cases using `System.getProperty("useKsp")`

I think that can work! Will test this out first thing in the morning😊

oas004:
@vinaygaba Can we re-run the UI Tests? It didn't provide me a log to check if it worked or not 😅

.github/workflows/android.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           api-level: ${{ matrix.api-level }}
           target: ${{ matrix.target }}
           arch: x86_64
-          script: ./gradlew connectedCheck --no-daemon --stacktrace -PuseKsp=false
```
vinaygaba, line 59: oh I was suggesting adding `-DuseKsp` in addition to `-PuseKsp` - . The gradle files add dependencies based on the project flags so with the current setup, only kapt is running and ksp isn't even running in the tests - https://github.com/airbnb/Showkase/blob/master/sample/build.gradle#L67

sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
```diff
@@ -44,3 +47,24 @@ fun BasicChipYellowPreview() {
...
+@Composable
+fun BasicChipBluePreview() {
+    BasicChip(
+        text = "Chip Component",
+        modifier = Modifier.background(color = Color.Blue)
+    )
+}
+
+
+@ShowkaseComposable(name = "Basic Chip Grey Light", group = "Chips")
```
vinaygaba, line 62: Since the name and group are identical with the preview ones, I feel like these composable should be getting deduped 🤔 Is that not what you observed? 

You should only see 2 composables instead of 4 in the browser

.github/workflows/android.yml
oas004: Ah, sorry. Totally my bad. I misunderstood that. Added it back in e2d2d67 :) Does this look correct?

sample/src/main/java/com/airbnb/android/showkasesample/BasicChip.kt
oas004: I might be mistaken here, but I think they have different names from the Preview ones above at least. They have `Basic Chip Blue Light/Dark` and these have `Basic Chip Grey Light/Dark` 🤔 Then I think 4 components should be correct.  I could add a test case to see that they group as well if they have same name :) Wdyt? :)

vinaygaba: @oas004 Clearly I'm getting old. Sorry I misread that. You are correct!

.github/workflows/android.yml
vinaygaba: Perfect!

showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
```diff
@@ -37,31 +36,54 @@ class ShowcaseBrowserTest {
             }
         )
 
+    // This will alter now since KSP supports stacked preview annotations and KAPT does not.
+    private val componentSize = if (System.getProperty("useKsp") == "true") {
```
vinaygaba, line 40: Can you add a more descriptive comment about why we are doing this. Along with link to the issue

vinaygaba (approved):
Very exciting! Thank you for your patience and getting this to the finish line 👏

showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
oas004: Good idea! Added something in [fb87f3c](https://github.com/airbnb/Showkase/pull/259/commits/fb87f3cf14301ee3f72b288ce9955b59b0d32b1b) :) Does that look okay? :)

oas004:
> Very exciting! Thank you for your patience and getting this to the finish line 👏

This is supercool! I will start on the actual mutlipreview annotation as soon as possible! 😄 🤩  Thank you so much for your valuable input! 🙏  Feel like I am learning a lot from this! 😃 😍

showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
```diff
@@ -544,4 +648,34 @@ class ShowcaseBrowserTest {
             verifyButtonWithTagIsDisplayedAndEnabled("SearchIcon")
         }
     }
+
+    @Test
+    fun stacked_preview_show_up_in_browser() {
+        // Stacked previews are only supported from ksp, so this is to bypass kapt on CI
+        if (System.getProperty("useKsp") == "true") {
```
vinaygaba, line 655: Wouldn't you need to use `if(BuildConfig.IS_RUNNING_KSP)` here instead of System.getProperty?

oas004: Oh no! Yes you are totally correct! I guess they where just not running on the new test then 🤔 Changed it up in [f2003ad](https://github.com/airbnb/Showkase/pull/259/commits/f2003ad60499ea65cd6e1ed304de93392f6770f6) Thanks for catching that! :)

vinaygaba (approved):
Looks like we have everything covered at this point 🥳

Issue #250, opened 2022-08-05
Handle private @Preview annotations gracefully


micHar:
When trying to build a project with private Previews, Showkase complains:
```
The methods annotated with Preview can't be private as Showkase won't be able to access them otherwise
```

Does it have to be an error? Not all previews are Showkase-worthy, some of them can be used just in ide. Maybe it could be a warning or there could be a gradle flag for this? Wdyt? This should be relatively easy, I could contribute if you decide that it's a good idea to add it.

micHar:
Tbh, I would go as far as to include an option to opt out of using `@Preview` annotations altogether. If I can achieve all the same things with a dedicated showkase annotation, it makes the intent more explicit and is more future-proof as custom annotations will inevitably have more parameters than @Preview.

vinaygaba:
@micHar Actually, there is an option that's meant to do exactly this. Use the `skip` property in `@ShowkaseComposable`. Essentially, the preview will look like this.

```kotlin
@ShowkaseComposable(skip=true)
@Preview
private fun MyPrivateComposable() {

}
```

micHar:
Hmm, I'm not a big fan of this solution. It's like the library hijacks the more general annotation for its specific purpose. Would be more elegant to at least ignore the private preview and not crash the build when lib can be graceful about it.

vinaygaba:
@micHar The reason this was done is because silently ignoring would be a lot worse since the expectation for most new users would be that this just works out of the box. Maybe I can instead make this a warning so that at least the user is notified in some way.

micHar:
Yes, seems like a better idea to warn than to crash. Generally I would agree with you when it comes to the fail fast principle, but here it feels like showkase should be an addition and not hijack the original process. 
Alternatively maybe a compiler flag or sth to disable this behavior so that we don't have to use showkase annotations on composables which are not meant for showkase at all.

iamutkarshtiwari:
I have all my preview functions public (neither internal nor private) but I am still getting this error with `1.0.0-beta14`

iamutkarshtiwari:
@vinaygaba I'd also be nice to throw the lines in the error log at which the `@Preview` annotated methods are private. 
This was we can quickly navigate to that specific erroneous method and fix the build issue.

I was trying to integrate showkase in my project which has around 100s of `@Preview` annotated methods and showkase was throwing this aforementioned error. It took me a lot of time to find the ones that were actually failing the build.

vinaygaba:
[...]

PR #261, opened 2022-09-06
Allow private composable to be skipped in case of failing on build and handle them gracefully

vinaygaba:
Fixes https://github.com/airbnb/Showkase/issues/250


sample/build.gradle
```diff
@@ -57,6 +57,18 @@ android {
         exclude 'META-INF/gradle/incremental.annotation.processors'
     }
 }
+if (project.hasProperty("useKsp")) {
+    ksp {
```
vinaygaba, line 61: I think you might be able to take some inspiration from this PR. I have a mechanism to pass data to the annotation processor. You will need something similar for custom multi preview annotations @oas004

oas004: Yes indeed! Thank you for this! This is very nice!😊😍 I think it is going to be very similar @vinaygaba  😊

sample/build.gradle
```diff
@@ -57,6 +57,18 @@ android {
...
+        arg("skipPrivate", "true")
```
vinaygaba, line 62: is there a better naming convention I should follow for annotation processor options that are passed by the user?

showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/BaseProcessorTest.kt
```diff
@@ -3,10 +3,7 @@ package com.airbnb.android.showkase_processor_testing
 import com.airbnb.android.showkase.processor.ShowkaseProcessor
 import com.airbnb.android.showkase.processor.ShowkaseProcessorProvider
 import com.google.common.io.Resources
-import com.tschuchort.compiletesting.KotlinCompilation
-import com.tschuchort.compiletesting.SourceFile
-import com.tschuchort.compiletesting.kspSourcesDir
-import com.tschuchort.compiletesting.symbolProcessorProviders
+import com.tschuchort.compiletesting.*
```
vinaygaba, line 6: will remove

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
```diff
@@ -56,7 +57,10 @@ internal class ShowkaseValidator {
                 throw ShowkaseProcessorException(
                     "The methods annotated with " +
                             "$annotationName can't be private as Showkase won't be able to access " +
-                            "them otherwise.",
+                            "them otherwise. If you'd like to skip this check and ignore the private " +
+                            "previews, kindly pass skipPrivate=true as an annotation processor option." +
+                            "To learn more about how to set this option, read the Showkase README here- " +
+                            "https://github.com/airbnb/Showkase/blob/master/README.md",
```
pmecho, line 63: Reminder to add a section to the readme and maybe link directly to the section. I'm not seeing an existing section.

pmecho (approved):

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
vinaygaba: I'll only add it after I'm ready for the next release otherwise it will cause confusion

thereallukesimpson:
@vinaygaba Hi, thanks very much for this feature. Do you expect it to be released in the near future?

Issue #275, opened 2022-10-23
Unresolved Reference for GetBrowserIntent


Lotus-Flower:
I am getting an unresolved reference for `Showkase.GetBrowserIntent(context)` in the most recent version: `1.0.0-beta14`. This is using a bare-bones project with one module and two files: MainActivity and ShowkaseRootModule. Cleaning and rebuilding the project/ invalidating caches does not work. However, downgrading to `1.0.0-beta13` immediately fixes the issue.

Lotus-Flower:
Looks like this might be something with AS

Issue #265, opened 2022-09-12
Unresolved reference: getBrowserIntent


hamedsilver:
I'am adding showkase-processor with KSP 
but get Unresolved reference: getBrowserIntent

I'm using Showkase In multi module way:

app module:
     * just use Theme and start showCase Activity*
        setContent {
            startActivity(Showkase.getBrowserIntent(this))
            finish()
        }

ui_core module(android Library module):
     *contain Compose components and Theme*

in both module I added the show Case dependencies
I check the generated file's in app module and find RootModuleShowkaseExtensionFunctionsCodegen

I used the implementation of the function directly inn my code for starting the Activity and it works.

    val intent = Intent(context, ShowkaseBrowserActivity::class.java)
    intent.putExtra("SHOWKASE_ROOT_MODULE", "com.mydigipay.designsystem.RootModule")

zaaaach:
I am getting the exact same thing using KSP.

vinaygaba:
Are you trying to reference the `getBrowserIntent` in the same module that you declared the `@ShowkaseRoot` class in?

```
@ShowkaseRoot
class MyRootModule: ShowkaseRootModule
```

zaaaach:
> Are you trying to reference the `getBrowserIntent` in the same module that you declared the `@ShowkaseRoot` class in?
> 
> ```
> @ShowkaseRoot
> class MyRootModule: ShowkaseRootModule
> ```

Yes. I added an identical `RootModule.kt` as in the sample. I can also see the code is being generated in the `build` folder and `getBrowserIntent` exists.

vinaygaba:
If the file is correctly generated with the `getBrowserIntent`, there's no reason for you to not be able to access it, unless you are trying to access it from a different package. Can you show me what the generated file that contains the `getBrowserIntent` looks like?

zaaaach:
Well, that's what odd to me too.

When I search I see multiple files containing `getBrowserIntent` one for each build flavor all inside of app/build/generated/ksp/{buildFlavor}/kotlin/{packageName}/RootModuleShowkaseExtensionFunctionsCodeGen.kt

Here's the contents of the file: 
`// This is an auto-generated file. Please do not edit/modify this file.
package <packagename>

import android.content.Context
import android.content.Intent
import com.airbnb.android.showkase.models.Showkase
import com.airbnb.android.showkase.models.ShowkaseElementsMetadata
import com.airbnb.android.showkase.models.ShowkaseProvider
[...]

PR #279, opened 2022-11-25
Let the IDE recognize generated code when using KSP

### LINKED ISSUE #278: Showkase#getMetadata() is not generated
The main question is how to get this generated?

> The Showkase object is the receiver for all the helper methods that this library generates. Currently there are a few extension functions that are generated with the Showkase object as the receiver. In order to get access to these functions, you need to build the app once so that the methods can be generated for your use.

From the doc here, I have to build the app in order to do so. But what if I'm looking for @Preview components from n modules/libraries and there is no such things as an app in my libraries. For example, think of a design system library, there is no app, unless we build like an example app. How can I get that Showkase#getMetadata() generated?

Related: [link](https://github.com/airbnb/Showkase/pull/252#issuecomment-1312462298). I'm trying to do similar thing here!

Omico:
Fix #278, #265, #275
See also: https://kotlinlang.org/docs/ksp-quickstart.html#make-ide-aware-of-generated-code

Extra note for someone using android variants, you should use `androidComponents.onVariants()` instead.


vinaygaba (commented):
Thanks for making this change. I had actually attempted to fix this here for some tests when I found a similar issue - https://github.com/airbnb/Showkase/blob/master/showkase-screenshot-testing-paparazzi-sample/build.gradle#L79

I didn't realize this was happening for all the cases. Can you also clean up the code I pointed to? After the changes you are adding, can you check what would still be needed in that file?

vinaygaba (approved):
Thanks for this change!

PR #281, opened 2022-12-05
Upgrade the release process in line with vannitech library upgrade so that we can release again.

vinaygaba:


publishing.gradle
```diff
@@ -0,0 +1,26 @@
...
+        }
+    }
+}
+
+mavenPublishing {
+    if (findProperty("doNotSignRelease").toString().toBoolean()) {
+        println("Skipping release signing")
+    } else {
+        println("Signing release with gpg")
+//        signAllPublications()
```
oas004, line 24: Should it not be signed in this case? Or why do we need to have this commented out? :)

oas004 (approved):
LGTM! 🙌

Issue #274, opened 2022-10-13
add support to other build types than only debug and release 


kibotu:
Currently only debug and release build types are supported

imagine having a setup with build types

```groovy

android {

    buildTypes {

        alpha {
            initWith(buildTypes.debug)
        }

        release {
        }

        beta {
            initWith(buildTypes.release)
        }
    }

    dependencies {
        alphaImplementation "com.airbnb.android:showkase:1.0.0-beta14"
        kaptAlpha "com.airbnb.android:showkase-processor:1.0.0-beta14"
    }
}

```

then the dependencies won't be available in alpha builds because the way it is published to maven central 

https://repo1.maven.org/maven2/com/airbnb/android/showkase/1.0.0-beta14/

compared to e.g. leakcanary 

https://repo1.maven.org/maven2/com/squareup/leakcanary/leakcanary-android/2.9.1/

the problem lies with the file naming 

where instead of having one artifact, the showkase library has 2: one for debug, which is not a bad idea if you want automatic switching with just adding implementation instead of having to add **debugImplementation** and **releaseImplementation**

however I'm not aware of a way to have those files added for alpha/beta/release or any other non default build type now

--- 

for this to work we would have to have two different artifacts, e.g. one for 

```groovy
    com.airbnb.android:showkase:1.0.0-beta14 
    com.airbnb.android:showkase-no-op:1.0.0-beta14 // e.g. for release 
```

kibotu:
ok turns out there is a workaround

```groovy
android {

    buildTypes {

        alpha {
            matchingFallbacks = ['debug']
        }
    }
}
            
```

popovanton0:
I also encountered this problem, spent a whole day on it, and even wrote a [Medium article](https://medium.com/@popovanton0/fixing-dependency-metadata-in-gradle-65fd86abe4e1) about solving it.

TLDR; As it turns out, if you have a project like this:
```mermaid
graph LR
  B(base-ui) --> D("design-system<br>(with showkase)")
  A(app) --> A1(feature-module) --> B
  A1 --> D
```
you need (not entirely sure why) to specify `matchingFallbacks` in all `build.gradle` files. I did this using a root file that applies to all modules the project.

Unfortunately, I haven’t realized that need in time and wrote _almost_ a [Gradle plugin](https://gist.github.com/popovanton0/f73631c6a819c630d543c10f8e42fff3#file-gradlemetadatarule-gradle-kts) to modify dependency metadata.

[...]

PR #291, opened 2022-12-20
Stop publishing separate debug and release artifacts for the "showakse" artifact on maven

vinaygaba:
Fixes https://github.com/airbnb/Showkase/issues/274

Showkase was publishing separate `debug` and `release` artifacts to maven and this was causing weirdness when consumers had custom build types that weren't `debug` and `release`. Moreover, this wasn't intended to be the case in the first place so I'm attempting to fix this for the future releases. 

Screen Shot 2022-12-20 at 3 03 35 PM

### DIFF
--- a/showkase/build.gradle
+++ b/showkase/build.gradle
@@ -14,13 +14,6 @@ android {
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
     }
-
-    buildTypes {
-        debug {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
-        }
-    }
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_11
         targetCompatibility JavaVersion.VERSION_11


showkase/build.gradle
```diff
@@ -14,13 +14,6 @@ android {
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
     }
-
-    buildTypes {
```
vinaygaba, line 18: I'm not 100% sure that this is what was causing the issue but think this might have been the cause...

vinaygaba:
Okay I've found the root cause. It seems like bumping up the version of the `gradle-maven-publish-plugin` to 0.19 caused a behavior change. - https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md#version-0190-2022-02-26

Screen Shot 2022-12-20 at 3 35 10 PM

bryantJB:
As of 1.0.5 it seems that this issue may have come back? We use a "mock" build type that was working with 1.0.4
A snippet from the error:
 > - Variant 'debugVariantMaven***Publication' declares a library:
      - Incompatible because this component declares a component for use during compile-time, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug' and the consumer needed a component for use during runtime, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'mock'
      
This also gives errors for `releaseVariantMaven***Publication` 
with *** being JavaDoc, Runtime, Source, and Api
This leads me to believe that showkase is once again being published with `debug` and `release` variants

PR #300, opened 2023-02-06
Set update processor test output flag to false

oas004:
Noticed this when working on adding processor tests to custom preview annotation feature that this flag was set to true. 

I might be mistaken here, but I think this will then generate the output for all the tests and not actually verify the test themself. Might have been a miss in a feature or something. 

I think also it can be a bit mind bothering for people coming into the project to see that their test output is updating every time they run the test without taking the action to actually set the flag locally themself.

However, if you don't agree, feel free to close this PR.


vinaygaba (approved):
Wow thank you for catching this! Certainly wasn't intended.

oas004:
From the last failure, it says that it can't apply the KSP plugin for the version 1.7.10-1.0.6, but only for api 29 🤔 That is a bit strange.

oas004:
Rebasing with latest master ☝️

vinaygaba:
@oas004 I merged this PR!

PR #302, opened 2023-02-08
Upgrade Paparazzi to 1.2.0

vinaygaba:
Also changed rendering mode of Paparazzi to only occupy enough space for components as opposed to taking the entire screen + hiding system UI

@airbnb/showkase-maintainers


yvonnewong (approved):

oas004 (approved):
Nice!

PR #263, opened 2022-09-09
Add support for Custom Preview Annotations

oas004:
## Add support for CustomMultiPreviewAnnotations in Showkase.

This is the third PR in a series to add support for Custom MultiPreview Annotations. Referencing as a followup on #259 and #255.

By completing this should solve #233 

## What do I mean by Custom MultiPreview Annotations?

This is an annotation in the form 

```kt
@Preview(name = "Custom Preview One First", group = "Custom Previews")
@Preview(name = "Custom Preview One Second", group = "Custom Previews")
annotation class CustomPreview
```

this can be used to annotate a composable function like you would do any other preview function in Compose like this:

```kt
@CustomPreview
@Composable
fun CustomAnnotationPreview() {
}
```

This can also be combined like this:

```kt
@Preview(name = "Custom Preview One First", group = "Custom Previews")
@Preview(name = "Custom Preview One Second", group = "Custom Previews")
annotation class CustomPreviewOne


@Preview(name = "Custom Preview Two First", group = "Custom Previews")
@Preview(name = "Custom Preview Two Second", group = "Custom Previews")
annotation class CustomPreviewTwo

@CustomPreviewOne
@CustomPreviewTwo
@Composable
fun CustomAnnotationPreviewCombined() {
}
```

They should be able to have the custom annotations and the use  of them in different modules in a project.

For more information, please see docs on tooling in compose [here](https://developer.android.com/jetpack/compose/tooling)

## Goal

The goal is to add support for this in Showkase. We would like showkase to act like the tooling in Android Studio does. That means that when Android Studio Preview would have generated a Preview, showkase should also generate a ShowkaseMetadata function.

### Sample App

In the sample app I have introduced `Shapes`. This is to showkase different shapes in Showkase.

Here I have made two internal (in same module) and one external (in submodule) annotations

Internals:

```kt
@Preview(name = "Square 200 by 200", group = "Shape", widthDp = 200, heightDp = 200)
@Preview(name = "Square 400 by 400", group = "Shape", widthDp = 400, heightDp = 400)
annotation class ShapePreview

@Preview(
    name = "Phone 200 by 200",
    group = "Shape",
    widthDp = 200,
    heightDp = 200,
    device = Devices.PHONE
)
@Preview(
    name = "Tablet 400 by 400",
    group = "Shape",
    widthDp = 400,
    heightDp = 400,
    device = Devices.TABLET
)
annotation class DevicePreview
```
Eksternal

```kt
[...]

PR #303, opened 2023-02-12
Add support for CustomPreviewAnnotations in Showkase 🚿 🌈 

### LINKED ISSUE #233: Support multi-preview annotations (annotations annotated with @Preview)

oas004:
This was moved from #263 to avoid a lot of the heat after rebasing. Also fixed some of the stuff for the compiler args so that it works for KAPT as well.

## Add support for CustomMultiPreviewAnnotations in Showkase.

This is the third PR in a series to add support for Custom MultiPreview Annotations. Referencing as a followup on #259 and #255.

By completing this should solve #233

resolves #233  

## What do I mean by Custom Preview Annotations?

This is an annotation in the form 

```kt
@Preview(name = "Custom Preview One First", group = "Custom Previews")
@Preview(name = "Custom Preview One Second", group = "Custom Previews")
annotation class CustomPreview1
```

```kt
@Preview(name = "Custom Preview One First", group = "Custom Previews")
annotation class CustomPreview2
```

this can be used to annotate a composable function like you would do any other preview function in Compose like this:

```kt
@CustomPreview1
@Composable
fun CustomAnnotationPreview() {
}
```

This can also be combined like this:

```kt
@Preview(name = "Custom Preview One First", group = "Custom Previews")
@Preview(name = "Custom Preview One Second", group = "Custom Previews")
annotation class CustomPreviewOne


@Preview(name = "Custom Preview Two First", group = "Custom Previews")
annotation class CustomPreviewTwo

@CustomPreviewOne
@CustomPreviewTwo
@Composable
fun CustomAnnotationPreviewCombined() {
}
```

They should be able to have the custom annotations and the use  of them in different modules in a project.

For more information, please see docs on tooling in compose [here](https://developer.android.com/jetpack/compose/tooling)

## Goal

The goal is to add support for this in Showkase. We would like showkase to act like the tooling in Android Studio does. That means that when Android Studio Preview would have generated a Preview, showkase should also generate a ShowkaseMetadata function.

## KAPT vs KSP

Because of https://youtrack.jetbrains.com/issue/KT-49682/Support-JVM-IR-in-KAPT-stub-generation KAPT does not quite support repeatable annotations from Kotlin. This means that  if you have an annotation class like:

```kt
@Preview(name = "Shape 100 by 100", group = "Shapes", widthDp = 100, heightDp = 100)
@Preview(name = "Shape 150 by 150", group = "Shapes", widthDp = 150, heightDp = 150)
annotation class CustomShape
```

This will be skipped by KAPT, but KSP will pick it up. However, if you have an annotation like:

```kt
@Preview(name = "Shape 100 by 100", group = "Shapes", widthDp = 100, heightDp = 100)
annotation class CustomShape
```
It will be picked up by both KSP and KAPT.

### Sample App
In the sample app I have introduced  CustomShape from the same module

```kt
@Preview(name = "Shape 100 by 100", group = "Shapes", widthDp = 100, heightDp = 100)
@Preview(name = "Shape 150 by 150", group = "Shapes", widthDp = 150, heightDp = 150)
annotation class CustomShape
```

And FontPreview from the same module

```kt
@Preview(name = "Normal font size", group = "FontPreview", fontScale = 1f)
@Preview(name = "High font size", group = "FontPreview", fontScale = 1.5f)
@Preview(name = "Super High font size", group = "FontPreview", fontScale = 2f)
annotation class FontPreview
```
You can check this in the sample app if you build it with KSP.

### Browser Test

In the browser tests I have introduced  CustomTextPreview

```kt

@Preview(
    name = "Custom Text Dark",
    group = "Custom Text",
    uiMode = Configuration.UI_MODE_NIGHT_YES,
)
annotation class CustomTextPreview

```

There was already a composable to verify compilation from a pervious PR, so used that as well.  Since the CustomTextPreview only had one Preview annotation, this was tested to be visible with KAPT as well.

I have done the same with the submodule in the browser test. However, here I introduced  CustomSizePreview and CustomFontSizePreview and verified that it works with both KAPT and KSP.

```

@Preview(name = "CustomSize 100 * 100", widthDp = 100, heightDp = 100, group = "CustomSubmodulePreview")
@Preview(name = "CustomSize 200 * 200", widthDp = 200, heightDp = 200, group = "CustomSubmodulePreview")
annotation class CustomSizePreview

@Preview(name = "Custom Font Size 1.2f", fontScale = 1.2f, group = "Custom Size Submodule")
annotation class CustomFontSizePreview
```

### Processor Test

In the processor tests, I have made tests that make sure all the correct metadata objects are created and that we are creating the files to store the custom annotations that the processor has registered. You can check these tests in `showcase-processor-testing` module.

 - [x] Make custom multi preview annotations work for annotations that are in the same module as the showkase preview.
 - [x] Make custom multi preview annotations work for annotations that are in a different module as the showkase preview and is imported into the file that has the showkase preview.
 - [x] Add processor tests for the different cases.
 - [x] Add Browser tests for the singular module case.
 - [x] Add Browser tests for the multi module case.
 - [x] Add processor test for combining custom annotations.
 - [x] Add processor tests for KAPT case and KSP case.


anhanh11001:
Thanks a lot for this work @oas004 🙏 🙏 🎆 🍾

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -82,7 +102,8 @@ class ShowkaseProcessor @JvmOverloads constructor(
     private fun processComponentAnnotation(roundEnvironment: XRoundEnv): Set<ShowkaseMetadata.Component> {
         val showkaseComposablesMetadata = processShowkaseAnnotation(roundEnvironment)
         val previewComposablesMetadata = processPreviewAnnotation(roundEnvironment)
-        return (showkaseComposablesMetadata + previewComposablesMetadata)
+        val customPreviewFromClassPathMetadata = processCustomAnnotationFromClasspath(roundEnvironment)
```
vinaygaba, line 105: why do we look at the classpath already? we want to pick up from classpath when we want the metadata generated from the other module. At this point of time, we want to process just the current module. Is this just a naming thing?

vinaygaba: shouldn't you call `processCustomAnnotation` here instead? 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +155,93 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+        val supportedTypes = mutableListOf<String>()
+        if (annotation != null) supportedTypes.add(annotation.qualifiedName)
+        supportedTypes.addAll(supportedCustomAnnotationTypes())
+        val components = mutableSetOf<ShowkaseMetadata.Component>()
+
+        supportedTypes.map { supportedType ->
+            val annotatedElements = roundEnvironment.getElementsAnnotatedWith(supportedType)
+            annotatedElements.map { annotatedElement ->
+                if (showkaseValidator.checkElementIsAnnotationClass(annotatedElement)) {
+                    processCustomAnnotation(roundEnvironment, annotatedElement)
```
vinaygaba, line 171: Why do you need to call this recursively? 🤔 I didn't follow this part

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +155,93 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+                            ShowkaseBrowserWriter(environment).writeCustomAnnotationElementToMetadata(
+                                xElement
+                            )
+                            return@elementRoot null
+                        }
+                        showkaseValidator.validateComponentElementOrSkip(
+                            xElement,
+                            customPreviewMetadata.supportTypeQualifiedName,
+                        )
+                        components.add(
```
vinaygaba, line 230: you don't need to add it to a set like this. Just collect the whole thing in a variable. For example

```kotlin
return supportedCustomPreview
            .mapIndexed {...}.toSet()
```

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -50,17 +52,35 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
 
     override fun getSupportedOptions(): MutableSet<String> {
-        return mutableSetOf("skipPrivatePreviews")
+        val supportedOptions = mutableSetOf<String>()
+        supportedOptions.add("skipPrivatePreviews")
+        supportedOptions.add("multiPreviewTypes")
+        return supportedOptions
+    }
+
+    // Getting the custom annotations that are supported as an compiler argument
```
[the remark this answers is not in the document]
oas004: Good idea! Added that in [d40c9d5](https://github.com/airbnb/Showkase/pull/303/commits/d40c9d5294899f92367fe0de61095c5336fecad4)  :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Yeah you are correct. Updated this in [d40c9d5](https://github.com/airbnb/Showkase/pull/303/commits/d40c9d5294899f92367fe0de61095c5336fecad4) :) Thanks for pointing it out! :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Initially I thought that if the element was a preview annotation I would like to call this to get the element that it is annotation. However, when thinking about it, I realise now that that will be picked up by when we check for preview annotations. Updated this in [d40c9d5](https://github.com/airbnb/Showkase/pull/303/commits/d40c9d5294899f92367fe0de61095c5336fecad4) :) Thanks!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: I think we can do it like that, it was just that since it had two mapIndexed method, I found this was a bit easier to read. I can ofc change it if you think your way is easier :) Wdyt? :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -50,17 +52,31 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+    }
+    override fun getSupportedOptions() = mutableSetOf("skipPrivatePreviews", "multiPreviewTypes")
+
+    // Getting the custom annotations that are supported as an compiler argument.
+    // It is expected to get the compiler argument as follows:
+    // arg("multiPreviewTypes", "com.airbnb.android.submodule.showkasesample.FontPreview")
+    private fun supportedCustomAnnotationTypes(): MutableSet<String> {
+        val set = mutableSetOf<String>()
+        environment
+            .options["multiPreviewTypes"]
```
vinaygaba, line 76: Should we add any kind of validation here?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -82,7 +98,8 @@ class ShowkaseProcessor @JvmOverloads constructor(
     private fun processComponentAnnotation(roundEnvironment: XRoundEnv): Set<ShowkaseMetadata.Component> {
         val showkaseComposablesMetadata = processShowkaseAnnotation(roundEnvironment)
         val previewComposablesMetadata = processPreviewAnnotation(roundEnvironment)
-        return (showkaseComposablesMetadata + previewComposablesMetadata)
+        val customPreviewFromClassPathMetadata = processCustomAnnotation(roundEnvironment)
```
vinaygaba, line 101: Maybe rename to 

```suggestion
        val customAnnotationComposablesMetadata = processCustomAnnotation(roundEnvironment)
```

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -112,7 +129,14 @@ class ShowkaseProcessor @JvmOverloads constructor(
         val skipPrivatePreviews = environment.options["skipPrivatePreviews"] == "true"
         return roundEnvironment.getElementsAnnotatedWith(PREVIEW_CLASS_NAME)
             .mapNotNull { element ->
-                if (showkaseValidator.checkElementIsAnnotationClass(element)) return@mapNotNull null
+                if (showkaseValidator.checkElementIsAnnotationClass(element)) {
```
vinaygaba, line 132: Since we are explicitly letting Showkase know which custom annotations are meant for preview, I think we can return null and skip them for processing, correct? 🤔

If this was working, then we wouldn't need the use annotation options at all. iirc, this was only working for ksp and not kapt, correct?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +151,93 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+        // In this function we are checking generated classpath for MultiPreview codegen annotations.
+        // We also check the current module if there is any composables that are annotated with the qualified name
+        // from the annotation from classpath. We use the fields from the classpath annotation to build
+        // common data for the ShowkasMetadata.
+
+        // Supported annotations from classpath
+        val supportedCustomPreview = environment.getTypeElementsFromPackage(CODEGEN_PACKAGE_NAME)
+            .flatMap { it.getEnclosedElements() }.mapNotNull {
+                when (
+                    val annotation = it.getAnnotation(ShowkaseMultiPreviewCodegenMetadata::class)
```
vinaygaba, line 199: if we are manually passing the custom annotation, then we don't need to create a separate codepath or generate any files with this custom annotation. We should be able to use the existing setup that `@Preview` and `@ShowkaseComposable` annotations are using. 

This approach was only needed if we were trying to infer the custom annotation. Is that what you are still hoping to do? Was the limitation that this was only working for ksp and not for kapt? I think we settled on just favoring consistency and going with the "manually specificying custom annotations" approach. At that point, we don't need this and should reuse the codepath used by `@Preview` and/or `@ShowkaseComposable`

vinaygaba: It's also possible I've forgotten some of the details we discussed earlier and I might be incorrect. Could you remind me?

vinaygaba: Hmm, actually there's the part where after identifying a custom annotation, we need to somehow "remember" the previews that it's annotated with and using that in the metadata. Where are you doing that part?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: I think when I tested the KAPT version of this without the compiler options, I tested it with repeatable annotations which do not work with KAPT anyways.  This was totally my mistake and might have lead to confusion.  I think we can ditch the compiler argument strategy and use this instead. IMO, this will be a better user experience for the enduser. Wdyt?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: When thinking about this, why can't we get annotations from class path at this point? What this method does it that it checks if there is any files that has a "saved" annotation with the `ShowkaseMultiPreviewCodegenMetadata` annotation so that we in this round can check for that annotation. The `processCustomAnnotation` will run when the preview check has found an element that is itself an annotation.

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: I think I also have forgotten some of the details that we discussed. I think we made the decision on if we should use the compiler argument approach on the claim that the `ShowkaseMultiPreviewCodegenMetadata` approach only worked with KSP. However, I think I tested that with repeatable preview annotations in KAPT. When testing it now, it works without the compiler args. IMO it would be a better approach for the end user if we did not go with the compiler arg approach. Wdyt?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: I don't quite get this. I think we have discussed this before @vinaygaba 🤔 Shouldn't we look at the classpath annotation to check the current module if we have used that annotation in the current module? Maybe I'm mistaken here.

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Yeah I agree. I removed this piece since it works with the codegen strategy :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
vinaygaba: Yeah this comment was assuming that we aren't trying to "infer" the annotations but the users are passing it manually. Since we are going the inference route, it can be reasonable to check for the classpath here

oas004: Aha, okay. Then I will keep it as this then :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -50,18 +52,19 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
-        ShowkaseColor::class.java.name,
-        ShowkaseTypography::class.java.name,
-        ShowkaseRoot::class.java.name,
-        ShowkaseScreenshot::class.java.name,
-    )
-
-    override fun getSupportedOptions(): MutableSet<String> {
-        return mutableSetOf("skipPrivatePreviews")
+    override fun getSupportedAnnotationTypes(): MutableSet<String> {
+        val supportedAnnotationTypes = mutableSetOf(
```
vinaygaba, line 56: you can just inline this instead of creating a separate variable

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +138,92 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+                    null -> null
+                    else -> annotation.value
+                }
+            }
+        val components = mutableSetOf<ShowkaseMetadata.Component>()
+
+        supportedCustomPreview
+            .mapIndexed { index: Int, customPreviewMetadata: ShowkaseMultiPreviewCodegenMetadata ->
+                roundEnvironment
+                    .getElementsAnnotatedWith(customPreviewMetadata.supportTypeQualifiedName)
```
vinaygaba, line 196: since we aren't registering this annotation in `getSupportedAnnotationTypes`, how can this work for kapt? 🤔 I think there should be a scenario where this wouldn't work

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
```diff
@@ -144,6 +159,48 @@ internal class ShowkaseBrowserWriter(private val environment: XProcessingEnv) {
...
+                        .build()
+                ).build()
+        }
+
+        fileBuilder.addType(
+            TypeSpec.classBuilder(generatedClassName).addFunctions(functions).build()
+        ).addFileComment("This is an auto-generated file. Please do not edit/modify this file.")
+        try {
+            fileBuilder.build().writeTo(environment.filer, mode = XFiler.Mode.Aggregating)
+        } catch (fileExists: FileAlreadyExistsException) {
```
vinaygaba, line 199: If you remove this, are you seeing an error/crash? I wonder if this is masking a potential actual issue in the implementation 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Hm, that is a good point 🤔 It works in the UI tests and in the sample apps for the main module and the sample module.  Maybe it should be tested with a third module? 🤔 Wdyt?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Good point! I will fix that! Thanks!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
oas004: I removed it and tested locally and did not see any difference, so removed it in 855cd00 :) Thanks for pointing it out! :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Fixed in 855cd00 :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Added registration on this to 855cd00, but I must say I'm wondering why this works without it and which case it does not 🤔

oas004:
Seems like the ui tests are crashing on sdk 26 default. For me it looks like the emulator is crashing before it starts the tests. Do u know how I can fix this @vinaygaba ? :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
vinaygaba: yeah I didn't understand this part either...

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +135,92 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+                    .mapIndexed elementRoot@{ elementIndex, xElement ->
+
+                        if (showkaseValidator.checkElementIsAnnotationClass(xElement)) {
```
vinaygaba, line 196: The `xElement` here is a custom annotation like `@CombinedPreview`. Under what circumstances would that annotation be annotated to an annotation class itself?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Are you sure that we really need to register annotation qualified names for it to work with KAPT? Is there any documentation on that? Do u know? :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: I think my thought with this was that if you had a case of something like 

```
@Preview
@CombinedPreview
fun SomePreview()
```
or 
```
@FontPreview
@CombinedPreview
fun SomePreview()
```

oas004: hmm, when thinking about it that is a good point! Maybe I can remove this check actually 🤔

oas004: Locally at least it worked without this check. I added another module as well to test it. Pushed the changes now :)

sample-submodule-2/src/main/java/com/airbnb/android/submodule/showkasesample/LocalePreview.kt
```diff
@@ -0,0 +1,6 @@
+package com.airbnb.android.submodule.two.showkasesample
+
+import androidx.compose.ui.tooling.preview.Preview
+
+@Preview(name = "English", group = "LocalePreview", locale = "en")
+annotation class LocalePreview
```
vinaygaba, line 6: Remove any custom preview annotations from this module. Instead, make this module have a dependency on `sample-submodule`.

Have a preview in `sample-submodule2` that is annotated by one of the custom annotations declared in a different module. I want to see how that setup would behave. Test it with both ksp and kapt.

oas004: Added a case for this now, and it works with KAPT if the compiler argument with the qualifierd name is provided :)

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -50,18 +52,33 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+        val supportedAnnotations = mutableSetOf(
+            ShowkaseComposable::class.java.name,
+            PREVIEW_CLASS_NAME,
+            ShowkaseColor::class.java.name,
+            ShowkaseTypography::class.java.name,
+            ShowkaseRoot::class.java.name,
+            ShowkaseScreenshot::class.java.name,
+            ShowkaseMultiPreviewCodegenMetadata::class.java.name,
+        )
+        supportedAnnotations.addAll(supportedCustomAnnotationTypes())
```
vinaygaba, line 65:
```suggestion
        // This is added specifically for kapt
        supportedAnnotations.addAll(supportedCustomAnnotationTypes())
```

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -127,6 +153,95 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
+            environment.getTypeElementsFromPackage(CODEGEN_PACKAGE_NAME)
+            .flatMap { it.getEnclosedElements() }.mapNotNull {
+                return@mapNotNull when (
+                    val annotation = it.getAnnotation(ShowkaseMultiPreviewCodegenMetadata::class)
+                ) {
+                    null -> {
+                        null
+                    }
+                    else -> {
+                        val codeGenAnnotation = ShowkaseMultiPreviewCodegenMetadata(
```
vinaygaba, line 207: am I reading this right? you are storing the value from the annotation in an annotation itself? 🤔 how does this work?

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
```diff
@@ -144,6 +159,44 @@ internal class ShowkaseBrowserWriter(private val environment: XProcessingEnv) {
         .addMember("numTypography = %L", typographySize)
         .build()
 
+    // This is to aggregate metadata for the custom annotation annotated with Preview
+    internal fun writeCustomAnnotationElementToMetadata(element: XElement) {
+        if (!element.isTypeElement()) return
+        if (element.isAnnotationClass() && element.qualifiedName == ShowkaseProcessor.PREVIEW_CLASS_NAME) return
```
vinaygaba, line 165: is there any circumstance where the element passed is `Preview`?

oas004: I hope not, maybe I can revert this check? 🤔

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oas004: Yeah this should be the same as returning `annotation` here I think. I think I can revert to that. My bad, thanks for catching it!

oas004: I think it has to be like this, because the annotation here is wrapped in an `XAnnotationBox`, so the value here is then the annotation

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
oas004: I think there is a case actually where this occurs. I think I have to keep this check.

vinaygaba (approved):
Fantastic job on this PR 👏🏻 This has been a long time coming but I'm so glad you were able to be persistent and took care of all the cases that we could think of.

anhanh11001:
@vinaygaba When this is merged, can you please also make a new release? This would help me with my work at work a lot! Thanks guys for this work 🙏

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/writer/ShowkaseBrowserWriter.kt
oas004: I think this is if you have a repeatable preview annotation. Then first when the top preview is discovered by the processor it will see that the element annotated with preview is a preview annotation and then try to run this function. But the information we want to write to the file is contained in the annotation in the next round. I think that is why we need this check somewhere in the process here or at call-site 🤔

oas004:
> Fantastic job on this PR 👏🏻 This has been a long time coming but I'm so glad you were able to be persistent and took care of all the cases that we could think of.

Wow this is awesome! 🥳 Thanks a lot for all the help and mentorship through the process! Learned a lot from this!

anhanh11001:
I see, I also use `ksp`. Then it could be something related to my setup. I'll double check. Thanks for confirming

oas004:
I think this should work out of the box if you are using KSP.   I will try to check this out as well!

takahirom:
I had the same problem and I found that we should apply ksp to the module we have the custom preview annotation.
https://github.com/DroidKaigi/conference-app-2023/pull/605/files#diff-deb2ae6119caad892229b16120b0d33fe23ff75d1430dfb235c463482f10c386R10

PR #305, opened 2023-03-06
Support annotation customizations

pmecho:
We have a use case for a customized annotation that can carry more information than the standard `@ShowkaseComposable`. To avoid needing to specify another annotation on preview functions that has that information, this adds a compiler flag to allow a custom annotation to be used instead of the standard annotation.

Additionally, there are specific terms we use for the fields in the annotation and we'd like to customize those as well. This adds compiler flags to allow for overriding the `name`, `group` and `styleName` fields. It also allows for multiple fields to be defined in a priority order where the first non-empty field is used. For us, this allows us to separate test-specific previews from previews used as a canonical style in our design system.


showkase-processor-testing/src/test/java/com/airbnb/android/showkase_processor_testing/AnnotationCustomizationsTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    fun `custom annotation with custom fields`() {
+        val options = mutableMapOf<String, String>()
+        options[ShowkaseComposable.AnnotationOverride] = "ShowkaseProcessorTest.custom_annotation_with_custom_fields.input.CustomAnnotation"
+        options[ShowkaseComposable.FieldOverrideName] = "label"
+        options[ShowkaseComposable.FieldOverrideGroup] = "category"
+        options[ShowkaseComposable.FieldOverrideStyleName] = "style"
+        // This is only supported in KSP at the moment
+        compileInputsAndVerifyOutputs(options = options, modes = listOf(Mode.KSP))
```
pmecho, line 23: I'm not really sure why the field overrides only work on KSP at the moment. I didn't have a use for kapt so I didn't spend much time investigating it.

vinaygaba (commented):
We synced offline and figured out an alternate approach that should hopefully make the logic a lot easier to incorporate.

PR #313, opened 2023-04-04
Handle private previews for Custom annotations

### LINKED ISSUE #312: Custom MutliPreview annotations don't respect skipPrivatePreviews
Issue #233 and associated resolution PR #303 fail to properly handle MultiPreview annotations.

While trying to configure this in my project, I have a few custom multi preview annotations that are added to some private functions. The KSP step complains stating that the methods can't be private and that if I want it to skip them, I must configure the ksp arguments, which I have done, so the message is actually in error and the multipreview on private methods should be skipped.

oas004:
The validation was not quite handled for the custom annotation processing steps and it had not passed the flag since it had a default argument. This caused the validation to hit the step to indicate to add the private preview flag for the user even though they had already added that.

resolves #312 

This was totally my mistake, I should have handled this in my initial PR. Sorry about that

### DIFF
--- a/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
+++ b/showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
@@ -817,8 +817,11 @@ class ShowcaseBrowserTest {
         }
     }
 
+    // We have enabled private preview compiler flag so this test
+    // should test that we can have private methods annotated with custom annotations
+    // and that they will not show in the application.
     @Test
-    fun customAnnotatedPrivateComposablesShouldNotShot() {
+    fun customAnnotatedPrivateComposablesShouldCompileButNotShow() {
 
         composeTestRule.apply {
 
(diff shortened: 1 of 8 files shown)


sachin2dehury (approved):

showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
```diff
@@ -816,4 +816,36 @@ class ShowcaseBrowserTest {
 
         }
     }
+
+    @Test
+    fun customAnnotatedPrivateComposablesShouldNotShot() {
```
vinaygaba, line 821: I was wondering what `ShouldNotShot` implies here?

vinaygaba (approved):

showkase-browser-testing/src/androidTest/java/com/airbnb/android/showkase_browser_testing/ShowkaseBrowserTest.kt
oas004: Ah yeah, good point! I think I misspelled to `ShouldNotShow` here, but I can change it and add a comment maybe? :)

PR #309, opened 2023-03-27
Initial support for tags and extra metadata on components

pmecho:
This starts the support for adding generic strings to tag and label components to help with filtering and custom categorization. Tags are added to the browser for filtering while metadata is only accessible from the generated `ShowkaseCodegenMetadata` object. A follow-up PR will add visual tag filtering to the built in browser.

PR #315, opened 2023-04-05
De-dupe with the fully qualified function name

pmecho:
This adds the fully qualified name (package + class name + function name) to the de-duplication strategy to support showkase components with the same group/component name/style name. This is useful when creating custom browsers where additional categorization is made via the new `tags` or `extraMetadata` fields that were added with https://github.com/airbnb/Showkase/pull/309.

### DIFF
--- a/showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
+++ b/showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
@@ -275,13 +275,13 @@ class ShowkaseProcessor @JvmOverloads constructor(
     }
 
     private fun Collection<ShowkaseMetadata.Component>.dedupeAndSort() = this.distinctBy {
-            // It's possible that a composable annotation is annotated with both Preview &
-            // ShowkaseComposable(especially if we add more functionality to Showkase and they diverge
-            // in the customizations that they offer). In that scenario, its important to dedupe the
-            // composables as they will be processed across both the rounds. We first ensure that
-            // only distict method's are passed onto the next round. We do this by deduping on
-            // the combination of packageName, the wrapper class when available(otherwise it
-            // will be null) & the methodName.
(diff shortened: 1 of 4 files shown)


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -290,16 +290,18 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
         .distinctBy {
-            // We also ensure that the component groupName and the component name are unique so 
-            // that they don't show up twice in the browser app.
+            // We also ensure that the component groupName and the component name are unique so
+            // that they don't show up twice in the browser app. This also de-duplicates based
+            // on the fully qualified function name to support categorization with additional
+            // fields (e.g. tags, extraMetadata, etc) on custom browsers.
             if (it.componentIndex != null) {
-                "${it.showkaseName}_${it.showkaseGroup}_${it.showkaseStyleName}_${it.componentIndex}"
+                "${it.fqPrefix}_${it.showkaseName}_${it.showkaseGroup}_${it.showkaseStyleName}_${it.componentIndex}"
```
vinaygaba, line 298: we already do the `${it.packageName}_${it.enclosingClassName}_${it.elementName}` dedpuing above. Why do we need to do it again here 🤔

pmecho: @vinaygaba If I don't include it here as well it will remove ones that only differ by `tags` or `extraMetadata`. I tried to change this to do them in one `distinctBy` step but that messed up where a `@Preview` annotation has the same group/style as the `@ShowkaseComposable` annotation so I left it as two steps.

vinaygaba: I see!

vinaygaba (approved):
LGTM

PR #317, opened 2023-04-05
Update kotlin to 1.8.10, xprocessing to 1.6.0-alpha01

elihart:
Updates kotlin, ksp, and the xprocessing library.

Notably, the androidx.room:room-compiler-processing library (aka xprocessing) has been updated to 2.6.0-alpha01. This version is incompatible with previous versions due to a breaking API change. All annotation processors using this library must be on the same version. Other annotation processors such as Epoxy and Paris also use xprocessing and if you use them you need to use a version of them that also uses xprocessing 2.6.0-alpha01

For that reason we shouldn't merge this until all other planned work is merged and released so that it can be used without forcing updates of other libraries


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/BaseProcessor.kt
```diff
@@ -107,7 +106,11 @@ abstract class BaseProcessor(
         } catch (e: Throwable) {
             // Errors thrown from within KSP can get lost, making the root cause of an issue hidden.
             // This helps to surface all thrown errors.
-            messager.printMessage(Diagnostic.Kind.ERROR, e.stackTraceToString())
+            if (e is ShowkaseProcessorException && e.element != null) {
```
elihart, line 109: i noticed that the showkase logger in place doesn't every actually get used, so error messages haven't been logging the element. i added this to help, but it seems like there needs to be a bigger clean up

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/logging/ShowkaseValidator.kt
```diff
@@ -180,25 +188,34 @@ internal class ShowkaseValidator {
...
-                )
+
+        if (!element.isField()) {
+            throw ShowkaseProcessorException(
+                "Only \"Color\" fields can be annotated with $annotationName",
+                element
+            )
+        }
+
+        when (environment.backend) {
```
elihart, line 199: in KSP the type is now the color value class, not just Long

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -509,14 +507,8 @@ fun XElement.isTopLevel(enclosingElement: XMemberContainer): Boolean {
...
         val kotlinMetadata = xTypeElement.getFieldWithReflection<Any?>("kotlinMetadata")
-            ?: return true
 
-        val enclosingElementKind = kotlinMetadata
-            .getFieldWithReflection<KotlinClassMetadata.Class>("classMetadata")
-            .header
-            .kind
-
-        enclosingElementKind == FILE_FACADE_KIND
+        return kotlinMetadata == null
```
elihart, line 511: there is a new version of the metadata class that had breaking changes

pmecho (approved):

PR #294, opened 2022-12-24
Implemented Showkase + Paparazzi artifact to automate screenshot testing

vinaygaba:
This PR implements the `showkase-screenshot-testing-paparazzi` artifact that enables complete automation of screenshot testing for your codebase using Showkase + Paparazzi. 

Assuming you were already using Showkase and had previews of your `@Composable` functions (by using either `@Preview` or `@ShowkaseComposable`), this is all the extra setup you will need to add screenshot testing to your codebase using Showkase + Paparazzi

```kotlin
@ShowkaseScreenshot(rootShowkaseClass = YourShowkaseRootModuleClass::class)
abstract class MyPaparazziShowkaseScreenshotTest: PaparazziShowkaseScreenshotTest {
    companion object: PaparazziShowkaseScreenshotTest.CompanionObject
}
```

**With just 3 lines of code, all the previews from your codebase will be screenshot tested using Paparazzi with appropriate defaults** 🤯💥

You might have a use case where you also want to screenshot test your components in dark mode or in RTL. Doing so is equally trivial. Just implement the appropriate functions to override the default behavior.

```kotlin
@ShowkaseScreenshot(rootShowkaseClass = YourShowkaseRootModuleClass::class)
abstract class MyPaparazziShowkaseScreenshotTest: PaparazziShowkaseScreenshotTest {
    companion object: PaparazziShowkaseScreenshotTest.CompanionObject {
        override fun layoutDirections(): List<LayoutDirection> = listOf(LayoutDirection.Ltr, LayoutDirection.Rtl)

        override fun uiModes(): List<PaparazziShowkaseUIMode> = listOf(PaparazziShowkaseUIMode.DEFAULT, PaparazziShowkaseUIMode.DARK)
    }
}
```

This will now ensure that all the previews are also screenshot tested in RTL and dark mode configurations. This creates a test matrix so all permutations are screenshot tested, just like you'd expect. 

**Generated File**

```kotlin
// This is an auto-generated file. Please do not edit/modify this file.
package com.airbnb.android.showkase.screenshot.testing.paparazzi.sample

import androidx.compose.ui.unit.LayoutDirection
import app.cash.paparazzi.Paparazzi
import com.airbnb.android.showkase.models.Showkase
import com.airbnb.android.showkase.screenshot.testing.paparazzi.ColorPaparazziShowkaseTestPreview
import com.airbnb.android.showkase.screenshot.testing.paparazzi.ComponentPaparazziShowkaseTestPreview
[...]

PR #320, opened 2023-04-20
Add callback to control if screenshot should be taken

andersu:
Expanding on https://github.com/airbnb/Showkase/pull/294 by @vinaygaba I wanted more control of when the screenshot should actually be taken. 

My use case is that I want some screenshots to be taken for every locale that we support in our app. In these kind of previews we typically use localised string resources.
For other screenshots we just use dummy texts in English and only want one screenshot taken.

I have added one localised button preview where the text is "Button" in English and "Knapp" in Norwegian and I use my new callback in the sample module.

Let me know if you can think of other better ways to achieve what I want.

I notice that a lot of screenshots changed when I ran 
`./gradlew :showkase-screenshot-testing-paparazzi-sample:recordPaparazziDebug`
Let me know if you think I messed something up.


showkase-screenshot-testing-paparazzi-sample/src/test/java/com/airbnb/android/showkase/screenshot/testing/paparazzi/sample/MyPaparazziShowkaseScreenshotTest.kt
```diff
@@ -1,9 +1,23 @@
...
+import com.airbnb.android.showkase.screenshot.testing.paparazzi.PaparazziShowkaseDeviceConfig
 import com.airbnb.android.showkase.screenshot.testing.paparazzi.PaparazziShowkaseScreenshotTest
 
 @ShowkaseScreenshot(rootShowkaseClass = PaparazziSampleRootModule::class)
-abstract class MyPaparazziShowkaseScreenshotTest: PaparazziShowkaseScreenshotTest {
-    companion object: PaparazziShowkaseScreenshotTest.CompanionObject
+abstract class MyPaparazziShowkaseScreenshotTest : PaparazziShowkaseScreenshotTest {
+    companion object : PaparazziShowkaseScreenshotTest.CompanionObject {
+        override fun deviceConfigs(): List<PaparazziShowkaseDeviceConfig> {
+            return listOf("en-rGB", "nb-rNO").map { locale ->
```
vinaygaba, line 12: I'm still trying to wrap my head around your use case. If you only want to screenshot test `en-rGB`, why are you also passing the other locale here?

andersu: @vinaygaba thanks for checking.  

For some previews I want to take one screenshot for each locale and for others I want only one screenshot.
So when setting up the config for the screenshot tests I pass all the locales I want and then I use the group to mark some previews.

**Here's some examples from the app I work on:** 
1. Preview with localised strings. I want one for each of the 5 locales we support to see that the screen looks good in all languages.

Example:
```
@Preview
@Composable
fun NoResultsPreview() {
    NoResultsView(
        imagePainter = painterResource(id = R.drawable.img_no_results_stores),
        text = stringResource(id = R.string.store_search_no_results)
    )
}
```
image

2. Preview with English dummy string or without texts in it at all. If I still run it for every locale I get 5 screenshots that are exactly the same, so I only want one.

Examples: 
```
@Preview
@Composable
fun FullTopBarPreview() {
    TopBarView(
        title = "Top bar with icons",
        tintColor = White,
        backgroundColor = Orange,
        navigationButton = TopBarNavigationButton.BACK
    ) {
        TopBarIcon(TopBarMenuButton.SEARCH, tint = White)
        TopBarIcon(TopBarMenuButton.SCAN, tint = White)
    }
}
```
image

```
@Preview
@Composable
@ShowkaseComposable(group = "notLocalised")
fun QuantityControls99Preview() {
    QuantityControls(quantity = 99)
}
```
image

So my thought is I have to pass all the locales in the device configs and then have logic to skip some tests to avoid 5 screenshots of the same. In my project I do that with the group `notLocalised` and a setup close to Alex Vanyo's and this check: 
```
 // Only take one screenshot if the preview is not localised
 if (locale == "nl-rNL" || componentPreview.showkaseBrowserComponent.group != "notLocalised") {
     paparazzi.snapshot {
         ... 
     }
}
```      

But when I tried moving to your new simpler setup with generated code from https://github.com/airbnb/Showkase/pull/294 I lost the capability to do it this way. 
That's what I am trying to solve with this PR.

vinaygaba: @andersu The problem is that this change is very specific to your use case and it's not something I'm keen on adding to a library that's meant to be more generic. I think a use case like this is better suited for a custom test class. You can reuse a lot of the code by extending the `PaparazziShowkaseScreenshotTest` class.

andersu: @vinaygaba I still need access to the `group` when I'm overriding `takePaparazziSnapshot` 
I have updated the PR. Is this something we could add?

vinaygaba: @andersu Just exposing the group is again limiting as it's specifically for your use case. The lens we want to use is that of something that's more generic. I think exposing the `ShowkaseBrowserComponent` (and color and typography) objects might be more useful. The right solution might be to ensure that they all extend a common sealed class/interface so that we can expose it via the `PaparazziShowkaseTestPreview` interface. What do you think?

andersu: @vinaygaba I've pushed some changes. Is this what you had in mind?
I'm open to naming suggestions for the sealed interface too. Went with `ShowkaseBrowserElement` for now.

showkase-screenshot-testing-paparazzi-sample/src/test/java/com/airbnb/android/showkase/screenshot/testing/paparazzi/sample/MyPaparazziShowkaseScreenshotTest.kt
```diff
@@ -1,9 +1,36 @@
...
-abstract class MyPaparazziShowkaseScreenshotTest: PaparazziShowkaseScreenshotTest {
-    companion object: PaparazziShowkaseScreenshotTest.CompanionObject
+abstract class MyPaparazziShowkaseScreenshotTest : PaparazziShowkaseScreenshotTest {
+    override fun takePaparazziSnapshot(
+        paparazzi: Paparazzi,
+        testPreview: PaparazziShowkaseTestPreview,
+        direction: LayoutDirection,
+        mode: PaparazziShowkaseUIMode
+    ) {
+        if (paparazzi.resources.configuration.locales.get(0).toString() == "en-gb" ||
```
vinaygaba, line 20: Could you revert these changes. I'd like for this example to be the simplest version to make it easy for the users of the library. Good to verify that it can now support your use case though!

showkase-screenshot-testing-paparazzi/src/main/java/com/airbnb/android/showkase/screenshot/testing/paparazzi/PaparazziShowkaseScreenshotTest.kt
```diff
@@ -133,6 +134,8 @@ interface PaparazziShowkaseScreenshotTest {
 }
 
 interface PaparazziShowkaseTestPreview {
+    val showkaseBrowserElement: ShowkaseBrowserElement
```
vinaygaba, line 137: these changes LGTM!

showkase-screenshot-testing-paparazzi-sample/src/test/java/com/airbnb/android/showkase/screenshot/testing/paparazzi/sample/MyPaparazziShowkaseScreenshotTest.kt
andersu: Reverted the code specific to my use case

andersu:
@vinaygaba I can't see any details of why the ui-testing workflows fail. Are there any commands I can run locally to try to figure it out?

PR #336, opened 2023-07-07
Fix `width` incorrectly passed to `showkaseHeightDp` param in `ShowkaseMetadata`.

julioromano:
Most likely a typo I suppose.


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/models/ShowkaseMetadata.kt
```diff
@@ -353,7 +353,7 @@ internal fun getShowkaseMetadataFromCustomAnnotation(
             showkaseName = showkaseName,
             showkaseGroup = showkaseGroup,
             showkaseWidthDp = if (width == -1) null else width,
-            showkaseHeightDp = if (height == -1) null else width,
+            showkaseHeightDp = if (height == -1) null else height,
```
vinaygaba, line 356: 😢 Thank you for fixing it, typo indeed!

Issue #244, opened 2022-07-18
No preview found in a minified and shrinked build (Proguard)


tomjegan:
With direct reference [to this issue](https://github.com/airbnb/Showkase/issues/165) we are also experience this problem with builds where we have pro guard enabled.

As per [VincentJouanne](https://github.com/VincentJouanne)'s comments we also would like to enable showkase for our internal QA builds such that design/product can inspect the components in app.

We also want to shift the annotation processing away from our dev/debug build as the build times are too high for this tool to be valuable as is unfortunately (even with ksp enabled).

Is it possible to provide a set of rules which will stop all the functions required from from converted?

Have tried [StylianosGakis](https://github.com/StylianosGakis)'s suggestion but does not appear to work for us. Have also tried:

```
-keep class **.ShowkaseModuleCodegen { }

-keep,allowobfuscation @interface androidx.compose.ui.tooling.preview.Preview

-keep @androidx.compose.ui.tooling.preview.Preview class *
-keepclassmembers class * {
    @androidx.compose.ui.tooling.preview.Preview *;
}

-keepclassmembers class * {
    @com.airbnb.android.showkase.annotation.ShowkaseComposable <methods>;
}
```

But still seeing the error that there are no components.

Are we able to add in some default rules (or at least advise further in this instance) please?

vinaygaba:
@tomjegan Thank you for the tip 🥇

zaaaach:
@tomjegan Is the single `-keep class **.ShowkaseRootCodegen { }` all that is required? I have tried that line in combination alone and in combination in my app & module level proguard rules but I am still getting the no components error.

vinaygaba:
Just to summarize what's exactly needed for future reference - 

You essentially need to add a proguard keep rule to make it with in release/alternate builds. The exact keep rule depends on your setup. For example, if your `ShowkaseRootModule` implementation looks like this - 

```
package my.lovely.package

@ShowkaseRoot
class MyRootModule: ShowkaseRootModule
```

Then your proguard rule will need to be 

```
-keep class **.MyRootModuleCodegen { }
```

We are essentially trying to retain the final `XCodegen` class that Showkase generates. The name of this generated class is of the format `(_name_of_ShowkaseRootModule_implementation_)Codegen`. 

Hope this makes sense!

kotasjn:
@tomjegan Thank you for your insights! Works great! 🚀

Issue #199, opened 2021-12-09
The library doesn't work with Proguard/R8


davidvavra:
If the app has Proguard/R8 turned on, generated code is removed and Showkase doesn't show any components.

I workarounded it by adding an exception to Proguard rules - keep everything in a package with generated code - like this:

`-keep class com.jll.jet.app.showkase.** { *; }`

But probably more generic rule should be included in the library.

vinaygaba:
Are you using a release build of the Showkase browser? The primary use case of the app was in the debug mode so an R8 rule wasn't exported with the app. This is to ensure that we don't have an apk size impact on your app.

davidvavra:
Yes we have multiple flavors of the app and the use-case is that there is a flavor for QA environment. But we want to distribute release build so QA people test as close as production app possible. But seing Showkase is useful for QA people. Maybe R8 configuration can be only in documentation, not exported in the library.

lupsyn:
Any update on this ?

vinaygaba:
You essentially need to add a proguard keep rule to make it with in release/alternate builds. The exact keep rule depends on your setup. For example, if your `ShowkaseRootModule` implementation looks like this - 

```
package my.lovely.package

@ShowkaseRoot
class MyRootModule: ShowkaseRootModule
```

Then your proguard rule will need to be 

```
-keep class **.MyRootModuleCodegen { }
```

We are essentially trying to retain the final `XCodegen` class that Showkase generates. The name of this generated class is of the format `(_name_of_ShowkaseRootModule_implementation_)Codegen`. 

Hope this makes sense!

davidvavra:
@vinaygaba Thanks for the info. But it would be good to add this information to README before closing this issue.

PR #341, opened 2023-08-11
Embed proguard rule for generated provider class

mataku:
## Summary

Embed proguard rules for Codegen class so that minifyEnabled applications do not need additional rules.

Since the Codegen class extends the ShowkaseProvider class, additional rules are unnecessary for each application by providing rules for types that extend the ShowkaseProvider class.

consumerProguardRules is already specified in the project, so I added to it. 
https://github.com/airbnb/Showkase/blob/4a09e3ff1d3ef71c6b151b5c7deefc82529691c0/showkase/build.gradle#L15

refs. https://github.com/airbnb/Showkase/issues/199, https://github.com/airbnb/Showkase/issues/244

## Context

Currently, we need to add extra proguard rules if we set minifyEnabled as true

```kotlin
@ShowkaseRootModule
class SampleShowkaseRootModule : ShowkaseRootModule

// Showkase generated class
public class SampleShowkaseRootModuleCodegen : ShowkaseProvider
```

```shell
# application proguard rules
-keep class **.SampleShowkaseRootModuleCodegen { }
```

### DIFF
--- a/showkase/consumer-rules.pro
+++ b/showkase/consumer-rules.pro
@@ -1,0 +1,1 @@
+-keep public class * extends com.airbnb.android.showkase.models.ShowkaseProvider


vinaygaba (approved):
Thanks for the change!

mataku:
@vinaygaba 
There seems to be a problem with ui-testing workflow on GitHub Actions. I think macos-10.15 labelled runner is no longer available, so created a pull request to fix it: https://github.com/airbnb/Showkase/pull/342

Announcement: https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

Available Runners: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

mataku:
@vinaygaba 
Thanks for merging https://github.com/airbnb/Showkase/pull/342! Is there anything else I should do?

mataku:
The workflow execution seems to be unstable. The previous commit passed and the latest commit failed. I will look log details

vinaygaba:
@mataku yeah these builds have been flaky and I have not had the time to investigate. Any help on this front will be greatly appreciated 🙏🏻

mataku:
I think this pull request itself doesn't affect CI workflow's result because all projects in Showkase set minifyEnabled as false so generated apk doesn't apply R8 or proguard.

I investigated test results in my forked repo: https://github.com/mataku/Showkase/actions (WIP)

- At first it is rare to encounter failing tests
  - You may see action run count itself is too small in above link, but I retried many times in each runs. 
- encountered com.airbnb.android.showkase_browser_testing.ShowcaseBrowserTest#customAnnotatedPrivateComposablesShouldCompileButNotShow but no logs printed about it
  - https://github.com/mataku/Showkase/actions/runs/5934664335/attempts/33
- sample:connectedDebugAndroidTest often fails (but as noted above, in the first place, it is rare to encounter failing tests)

<details>
<summary> ShowcaseBrowserTest#customAnnotatedPrivateComposablesShouldCompileButNotShow results</summary>

ss- 2023-08-25 14 09 08

</summary>
</details>

mataku:
@vinaygaba 
How about treating instability in CI workflow as a new issue? If possible, I will create a new one.

For some time now, macOS specification on GitHub Actions has been out of date and the CI workflow has not been running, I don't understand the overall problem, but I think it was probably there originally as a potential issue.

In my investigation, the logs are printed unknown state, as if there is a system problem in the output or no logs at all in the first place, and the test will pass in many cases if it is retried. If tests related to changes in a pull request fail, the workflow will fail every time, so you can distinguish between the two.

mataku:
@vinaygaba 

> Additionally, even after removing the tests that were running on API 28

I think you still have left the (deleted) API 28 workflow set as "Require status checks to pass before merging"? Please check the Branch protection rules in Showkase repository settings

<details>
<summary>sample screenshot</summary>

screenshot 2023-08-27 19 37 34

</details>

vinaygaba:
@mataku I've reverted this PR. More explanation on why in the PR I updated it with - https://github.com/airbnb/Showkase/pull/354

Issue #347, opened 2023-08-25
exported ShowkaseBrowserActivity


Skeletonxf:
Hi, as part of an automated security scan on an app I'm working on Showkase was flagged due to the Showkase browser activity being exported by default.

| Issue  | Severity | Description |
| --- | --- | --- |
| Activity (com.airbnb.android.showkase.ui.ShowkaseBrowserActivity) is not Protected. [android:exported=true]  | high | An Activity is found to be shared with other apps on the device therefore leaving it accessible to any other application on the device.

This is easily overridden app side via

```xml
<activity
    android:name="com.airbnb.android.showkase.ui.ShowkaseBrowserActivity"
    android:exported="false"
    tools:replace="android:exported" >
</activity>
```

however the default being set to true could be surprising for some users of this library? I had initially thought that if I disabled the in app means of launching the browser then it wouldn't be accessible in a production build of the app, but given it is exported by default another malicious app could theoretically open this, and perhaps gain access to data in the screenshots that they shouldn't if anything sensitive is there.

rovkinmax:
I bealeve that ShowkaseBrowserActivity should not be included in a release build

vinaygaba:
Thanks for flagging. I see that this line was added in this PR (https://github.com/airbnb/Showkase/commit/469e14de06227d32974d2b3fa67646ff53013e17#diff-fdcbc654693654c9611addef6b4e7ef93e1ae16aa06522b1c02d9786a3d5ecb5R10) and I didn't catch it during the review. I've reached out to @oas004 to figure out why this was needed, otherwise I'm inclined to remove this line of code from the library.

vinaygaba:
Opened PR that fixes this - https://github.com/airbnb/Showkase/pull/355

PR #355, opened 2023-09-19
Stop exporting ShowkaseBrowserActivity

vinaygaba:
Fixes https://github.com/airbnb/Showkase/issues/347

This was done in this PR (https://github.com/airbnb/Showkase/commit/469e14de06227d32974d2b3fa67646ff53013e17#diff-fdcbc654693654c9611addef6b4e7ef93e1ae16aa06522b1c02d9786a3d5ecb5R10) in order to [make it compatible with Android 12](https://developer.android.com/about/versions/12/behavior-changes-12#exported). However, it's only needed for launcher activities (or activities with an intent filter), which `ShowkaseBrowserActivity` isn't. So it's safe to remove this line from our codebase.


oas004 (approved):
Looks good to me! Thanks for fixing this! I totally misread how exported should be used when we bumped the target

PR #360, opened 2023-10-11
Adding  generateScreenshot into ShowkaseComposable annotation 

IvanSanchez-uam:
The **generateScreenshot** property is used when showkase is integrated with paparazzi and the purpose of it is to filter all the previews that we want the screenshot test.

in the next code snippet code shows the usage of **generateScreenshot**

`// This preview will be skipped from component browser
// but will generate the screenshot
@ShowkaseComposable(
    name = "Basic Chip",
    group = "Chips",
    defaultStyle = true,
    generateScreenshot = true,
    skip = true
)
@Composable
fun BasicChipPreview() {
    BasicChip(text = "Chip Component")
}` 

**As you can see the screenshot test will be generated for the preview but will not be part of the component browser.**

The main idea of this pr is to have the ability to recognize previews that we want to make the screenshot test.


IvanSanchez-uam:
@marcorighini maybe you can help me out here, what do I need to get feedback for this PR?

PR #370, opened 2023-12-26
Update Showkase to latest version of Jetpack Compose Navigation 

designedbyz:
This PR updates the version of Jetpack Compose navigation to the latest version, 2.7.6, and updates other dependencies as needed along the way as well. This is needed to avoid some odd behavior with navigation 2.6.0; notably if you use compose navigation 2.6.0 and Showkase in the same project, the browser will not navigate from the root screen to any of typography, color, or composables. This seems to be an issue with Navigation 2.6.0, as both navigation 2.7.x and navigation 2.5.x work just fine. 

Other dependencies have been updated as needed as well, most notably perhaps was updating the compile SDK to version 34. This was required for navigation 2.7.x. 

I've seen a few issues while running tests on the project that I could use some help solving. Specifically, the `entering_text_in_search_bar_filters_the_visible_components` and `entering_text_in_search_bar_filters_the_visible_groups_of_components` are acting a bit weird detecting two nodes, even though I only see one visible node when debugging the tests. I'd appreciate some help with debugging this.


vinaygaba:
@designedbyz Genuinely grateful for your contribution. I wrote the browser piece of Showkase in 2020 when there was little to no documentation about Compose so there were a bunch of bad practices that I wanted to fix holistically. For example, there was no framework level solution to go back even 😅 I'm attempting to fix all that + the navigation dependencies in this PR - https://github.com/airbnb/Showkase/pull/390 🤞 

Would it be okay if I closed this PR as a lot needed to change for some of the bugs that I was observing to be fixed.

PR #372, opened 2024-01-11
Fix missing category of originating elements

vinaygaba:
Eagled eyed @elihart spotted that we were not adding the originating elements for previews that were without the `PreviewParameter`. This is the far more common case and could potentially be causing build issues when there were changes in the element. It primarily has implications on the ability to do proper incremental compilation


elihart (approved):

joshdmiller5:
Bumping @kmalmur ask here for a release of this, I believe it should solve: https://github.com/airbnb/Showkase/issues/373 as mentioned

Issue #340, opened 2023-08-09
Paparazzi + Showkase Integration - java.io.FileNotFoundException (File name too long)


ss0930:
I am using `showkase-screenshot-testing-paparazzi` from https://github.com/airbnb/Showkase/pull/294 to auto-generate snapshots for Composable Previews. For certain previews that have long names, I am running into this issue from Paparazzi https://github.com/cashapp/paparazzi/issues/734. Since the test name and png file name is being generated on Showkase, not sure if there is something I can do on my end.

I am able to get around it by renaming the Preview to something shorter, but would be great to have a fix for this in Showkase itself.

evant:
Ran into this as well, leaving some notes on how this can be improved:

1. It appears the generated file names for properties can be aggressively wrong, including the preview name, the component name, _and_ the group name concatenated together. beta17 was even worse including the package name as well but that appears to be fixed.

    Do we actually need all of those? I'd think the preview function name would be sufficient since you can't have duplicate function names in the source files anyway, but maybe there's something I'm missing there.

2. If you use preview params it also generates a lamba which is represented by a synthetic class. The name of this lambda is filename$propertyname$1.class which blows it up even more. shortening the filename/property would also help here but you can also go one step further and split it out into a separate function so the lamba isn't generated at all. Something like:
```kotlin
public val MyPreviewNameGroup: List<ShowkaseBrowserComponent> = 
    PreviewStyle()
        .values
        .iterator()
        .asSequence()
        .mapIndexed(::spacingPreviewSpacingExtrasSDKSpacingValues)
        .toList()
        
private fun myPreviewNameGroup(index: Int, previewParam: Param): ShowkaseBrowserComponent {
   return ShowkaseBrowserComponent(
        group = "Group
        componentName = "Name",
        componentKDoc =  "",
        componentKey = ...,
        isDefaultStyle = false,
        component = @Composable { MyPreview(param = previewParam) }
    )
}
```

vinaygaba:
@evant The reason all this info is added is because you can you repeated annotations and a single function can actually end up being multiple previews depending on how you are using the preview annotations. I do think that I need to figure out an alternate way to solve this though as hitting this issue could be common if you are using longer and more descriptive names. 

For example, here's a file that Paparazzi generated in the sample project --

```
[...]

PR #378, opened 2024-04-10
Generate files based on source file name instead of property name

evant:
This reduces the length of the filename because a disambiguated property name can be quite long. Looking at the files gerenated in the sample app, before this change the longest is 161 chars, after it's 102.

Partially addresses #340


evant:
Note: tests fail because of the filename changes, will update if this approach is accepted.

PR #385, opened 2024-05-25
Provide dark background for rendering Dark mode previews

### LINKED ISSUE #289: [Request] Provide dark background for rendering Dark mode previews
Apologies if this is already available, I have not seen any callouts for this in the README.

Basically, the request is to render a black card background in the `[Dark Mode]` preview generated by Showkase.

I defined my own Preview annotation which provides both the Light and Dark versions of content.
```kotlin
@Preview(
    name = "Light theme",
    showBackground = true,
    backgroundColor = 0xFFFFFFFF,
    uiMode = UI_MODE_NIGHT_NO
)
@Preview(
    name = "Dark theme",
    showBackground = true,
    backgroundColor = 0xFF000000,
    uiMode = UI_MODE_NIGHT_YES
)
internal annotation class PreviewDefaults
```

Which generates the following:

image

It would be great to have a similar functionality provided by Showkase out of the box. As it stands, the previews look like the following - note the content text in `[Dark Mode]` can't be read because the background is white, and not black.

image

As a workaround, I can wrap my preview content in a box that provides a background (with additional padding so we can see the dashed borders), but I do not think I should do this every time.

```kotlin
@ShowkaseComposable(
    name = PREVIEW_NAME,
    group = GROUP_CARD,
    defaultStyle = true,
)
@PreviewDefaults
@Composable
internal fun DashedCardPreview() {
    MyTheme {
        Box(
            modifier = Modifier
                .background(color = MyTheme.colors.background.page.default)
                .padding(16.dp)
        ) {
            DashedCardSample()
        }
    }
}
```

image

vinaygaba:
Fixes #289 

Notice the background of the dark mode card, specifically the area around the component itself. 

Screenshot 2024-05-24 at 9 17 38 PM


showkase/src/main/java/com/airbnb/android/showkase/ui/CommonComponents.kt
```diff
@@ -57,27 +60,34 @@ internal fun ComponentCardTitle(componentName: String) {
...
+    onClick: (() -> Unit)? = null,
+    darkMode: Boolean = false,
 ) {
     val composableModifier = Modifier.generateComposableModifier(metadata)
     val composableContainerModifier = Modifier.generateContainerModifier(onClick)
-    Card {
-        Box {
-            Column(modifier = composableModifier) {
-                metadata.component()
+    MaterialTheme(
```
vinaygaba, line 68: Ideally this was handled at the root screen itself but since I just realized Showkase wasn't really using the theme at all, I'm limit the scope and restrict it to just the dark mode use case.

PR #390, opened 2024-06-05
Improve ShowkaseBrowser codebase and update navigation stack

vinaygaba:


showkase/src/main/java/com/airbnb/android/showkase/ui/CommonComponents.kt
```diff
@@ -63,30 +67,35 @@ internal fun ComponentCard(
...
-    val composableModifier = Modifier.generateComposableModifier(metadata)
-    val composableContainerModifier = Modifier.generateContainerModifier(onClick)
-    MaterialTheme(
-        colors = if (darkMode) darkColors() else lightColors()
-    ) {
-        Card(
-            shape = MaterialTheme.shapes.large
+    // This is added to make sure that the navigation of the ShowkaseBrowser does not break
+    // when one of the previews has a back press handler in the implementation of the component.
+    val backPressedDispatcherOwner = rememberOnBackPressedDispatcherOwner()
```
vinaygaba, line 72: Doing this at the top level (which was previously the case) was causing the navigation to not work at all. Instead, I moved this down to specifically happen for the Composables that are responsible for rendering the component previews as that's where we want to override/intercept any back handling that the previews might be doing themselves.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseCategoriesScreen.kt
```diff
@@ -34,60 +37,62 @@ internal fun ShowkaseCategoriesScreen(
...
-                            ShowkaseCategory.COMPONENTS -> navController.navigate(
-                                ShowkaseCurrentScreen.COMPONENT_GROUPS
-                            )
-                            ShowkaseCategory.COLORS -> navController.navigate(
-                                ShowkaseCurrentScreen.COLOR_GROUPS
-                            )
-                            ShowkaseCategory.TYPOGRAPHY -> navController.navigate(
-                                ShowkaseCurrentScreen.TYPOGRAPHY_GROUPS
-                            )
+                            ShowkaseCategory.COMPONENTS -> onNavigateToComponentGroups()
```
eddnav, line 48: It would be nice if we could check here whether the component group only has one component and directly navigate into it.

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -436,103 +495,151 @@ private fun Map<String, List<*>>.isOnlyCategory(
...
-            showkaseBrowserScreenMetadata,
-            navController
+            groupedComponentMap = groupedComponentMap,
+            showkaseBrowserScreenMetadata = showkaseBrowserScreenMetadata,
+            onUpdateShowkaseBrowserScreenMetadata = onUpdateShowkaseBrowserScreenMetadata,
+            navigateTo = navigateTo,
         )
     }
     composable(ShowkaseCurrentScreen.COMPONENT_STYLES.name) {
         ShowkaseComponentStylesScreen(
```
eddnav, line 522: Similarly, skipping the styles screen when there's only one style would be super nice. Directly to the detail screen.

sample-submodule-2/build.gradle
```diff
@@ -61,7 +61,8 @@ dependencies {
     // Support Libraries
     implementation deps.support.appCompat
     implementation deps.support.ktx
-    implementation deps.support.lifecycleExtensions
+    implementation deps.support.lifecycleComposeViewModel
+implementation deps.support.lifecycleComposeRuntime
```
eddnav, line 65: Space was lost here :)?

sample-submodule/build.gradle
```diff
@@ -62,7 +62,8 @@ dependencies {
     // Support Libraries
     implementation deps.support.appCompat
     implementation deps.support.ktx
-    implementation deps.support.lifecycleExtensions
+    implementation deps.support.lifecycleComposeViewModel
+implementation deps.support.lifecycleComposeRuntime
```
eddnav, line 66: Same here

sample/build.gradle
```diff
@@ -94,7 +94,8 @@ dependencies {
     // Support Libraries
     implementation deps.support.appCompat
     implementation deps.support.ktx
-    implementation deps.support.lifecycleExtensions
+    implementation deps.support.lifecycleComposeViewModel
+implementation deps.support.lifecycleComposeRuntime
```
eddnav, line 98: And here

PR #401, opened 2024-12-11
Added ability to skip @Preview functions

vinaygaba:
This PR adds the ability to skip `@Preview` composable from being picked up by Showkase. I had previously added the ability to skip an individual composable using the [`skip` property](https://github.com/airbnb/Showkase/blob/master/showkase-annotation/src/main/java/com/airbnb/android/showkase/annotation/ShowkaseComposable.kt#L73) that's added to the `ShowkaseComposable` annotation. 

However, this config option allows one to skip all `@Preview` annotation with a single line. While Showkase picking up functions annotated with `@Preview` is an important feature that makes the library more useful, there might be cases where you might want to enforce usage of the `ShowkaseComposable` annotation instead as that's a more explicit way to  integrate with Showkase. In order to use this feature, you can use the following snippet in your gradle file.

If you use KSP:
```groovy
ksp {
 arg("requireShowkaseComposableAnnotation", "true")
}
```

If you use KAPT:
```groovy
kapt {
 arguments {
  arg("requireShowkaseComposableAnnotation", "true")
 }
}
```

This PR also contains upgrades to Detekt + Gradle and the fixes that detekt required as a result. Would've been ideal to split this in separate PR's but since I didn't anticipate needing these many changes and I was further along, I decided to just continue adding it to this PR. 

@elihart @allenchen1154


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -78,7 +78,7 @@ class ShowkaseProcessor @JvmOverloads constructor(
             ?.toSet()?.let { set.addAll(it) }
         return set
     }
-    override fun getSupportedOptions() = mutableSetOf("skipPrivatePreviews", "multiPreviewType")
+    override fun getSupportedOptions() = mutableSetOf("skipPrivatePreviews", "skipFunctionsWithPreviewAnnotation", "multiPreviewType")
```
vinaygaba, line 81: Very open to having a name other than `skipFunctionsWithPreviewAnnotation` 🫤

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -78,7 +78,11 @@ class ShowkaseProcessor @JvmOverloads constructor(
             ?.toSet()?.let { set.addAll(it) }
         return set
     }
-    override fun getSupportedOptions() = mutableSetOf("skipPrivatePreviews", "multiPreviewType")
+    override fun getSupportedOptions() = mutableSetOf(
+        "skipPrivatePreviews",
+        "skipFunctionsWithPreviewAnnotation",
```
elihart, line 83: might consider a more explicit name, as this won't skip _all_ functions with preview annotations (ie, it keeps the ones with showkase annotation)
```suggestion
        "skipPreviewFunctionsWithoutShowkaseAnnotation",
```

or 
```suggestion
        "requireShowkaseAnnotation",
```

elihart (approved):
Thanks a lot for adding this!

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
vinaygaba: this makes sense, thanks for the tip!

Issue #400, opened 2024-12-09
Provided Metadata instance has version 2.1.0, while maximum supported version is 2.0.0.


dostalleos:
After updating Kotlin to version 2.1.0, I can't build the project. I'm getting 
```
java.lang.IllegalArgumentException: Provided Metadata instance has version 2.1.0, 
while maximum supported version is 2.0.0. To support newer versions, update the kotlinx-metadata-jvm library.
``` 
error.

Showkase SDK: 1.0.3

Full stack trace:
```
error: java.lang.IllegalArgumentException: Provided Metadata instance has version 2.1.0, while maximum supported version is 2.0.0. To support newer versions, update the kotlinx-metadata-jvm library.
  	at androidx.room.jarjarred.kotlinx.metadata.jvm.KotlinClassMetadata$Companion.throwIfNotCompatible$kotlinx_metadata_jvm(KotlinClassMetadata.kt:609)
  	at androidx.room.jarjarred.kotlinx.metadata.jvm.KotlinClassMetadata$Companion.checkMetadataVersionForRead(KotlinClassMetadata.kt:600)
  	at androidx.room.jarjarred.kotlinx.metadata.jvm.KotlinClassMetadata$Companion.read(KotlinClassMetadata.kt:579)
  	at androidx.room.compiler.processing.javac.kotlin.KmClassContainer$Companion.createFor(KotlinClassMetadataUtils.kt:184)
  	at androidx.room.compiler.processing.javac.JavacTypeElement$kotlinMetadata$2.invoke(JavacTypeElement.kt:62)
  	at androidx.room.compiler.processing.javac.JavacTypeElement$kotlinMetadata$2.invoke(JavacTypeElement.kt:61)
  	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
  	at androidx.room.compiler.processing.javac.JavacTypeElement.getKotlinMetadata(JavacTypeElement.kt:61)
  	at androidx.room.compiler.processing.javac.JavacFieldElement$kotlinMetadata$2.invoke(JavacFieldElement.kt:54)
  	at androidx.room.compiler.processing.javac.JavacFieldElement$kotlinMetadata$2.invoke(JavacFieldElement.kt:53)
  	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
  	at androidx.room.compiler.processing.javac.JavacFieldElement.getKotlinMetadata(JavacFieldElement.kt:53)
[...]

PR #405, opened 2025-03-31
KSP2 / Room 2.7 support

oheyadam:
A lot of the work done here is based on changes made in [this CL](https://android-review.googlesource.com/c/platform/frameworks/support/+/3470790). 

I've also had to upgrade a few libraries for K2 compatibility.


showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
```diff
@@ -131,13 +134,21 @@ class ShowkaseProcessor @JvmOverloads constructor(
...
-        val skipPrivatePreviews = environment.options["skipPrivatePreviews"] == "true"
+        val skipPrivatePreviews = environment.options["skipPrivatePreviews"].toBoolean()
         val requireShowkaseComposableAnnotation =
-            environment.options["requireShowkaseComposableAnnotation"] == "true"
+            environment.options["requireShowkaseComposableAnnotation"].toBoolean()
 
         if (requireShowkaseComposableAnnotation) return emptySet()
 
         return roundEnvironment.getElementsAnnotatedWith(PREVIEW_CLASS_NAME)
+            .asSequence()
```
vinaygaba, line 144: For consistency, we should do this for all annotations (including ShowkaseComposable and custom annotations)

vinaygaba: Also maybe a comment explaining what it's trying to do as it's not obvious on first look for any future contributors.

vinaygaba:
@oheyadam The Paparazzi file renames are most likely due to the change in the sorting logic? but would be nice if you can confirm it. That and adding the sorting logic to the processing of other annotations would be the only thing that I think we'd need to add. Thank you for being an absolute legend and pushing through on this change completely independently. I'm grateful for your contribution and you will be helping a lot of others 🎖️

showkase-processor/src/main/java/com/airbnb/android/showkase/processor/ShowkaseProcessor.kt
oheyadam: Re: adding it to everything else, it seems like other methods aren't affected, or at least not as reliant on ordering as this one.

I just extracted this function and started adding it to other call-sites in this file where annotations are queried, and it doesn't seem to be affecting anything (as in the tests still pass), so I wonder if it would just be an unnecessary processing step? Happy to go either way, but just wondering it's worth it

iroyo:
Is there anything that still needs to be done? It seems the pipeline has failed more than a month ago.

dostalleos:
@vinaygaba @oheyadam Hello guys, when do you expect this PR will hit production? I'm currently unable to update the Kotlin version to 2.1.x in my project because of this. I created an issue: https://github.com/airbnb/Showkase/issues/400, and then found this PR. I believe it will solve my issue as well. Please make it available as soon as possible 🙏

vinaygaba (approved):

vinaygaba:
I was able to finally figure out why some of the tests were failing. One breaking change that I'll also add in the release notes is when a custom multi preview annotation is referenced from a different module than where the composable previews are declared. It feels like something changed in the KSP processing logic so if anywhere here can help me point to it, that would be fantastic. This is the commit I made to co-locate the previews and custom annotation declaration - https://github.com/airbnb/Showkase/pull/405/commits/10d455d89a73f01da1254f7b2c7b8d0f0dce4efd. Ideally this move wasn't needed (like before). 

Anyway, I think we should be unblocked for now and I'll do a release shortly 🎉

vinaygaba:
Thank you @oheyadam for doing the heavy lifting here and contributing to Showkase. God's work!!! 🙏🏻

PR #412, opened 2025-07-03
Support edge-to-edge content

### LINKED ISSUE #399: Android 15 edge-to-edge enforcement support
Hi,
Android 15 [has enforced](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) edge-to-edge automatically for all activities for apps targeting SDK 35.

Apparently, Showkase does not fully support it. 

If you look at the screenshot I provided, the activity is drawn behind the system bars and does not handle correctly the insets.

I had a look at the source code of the `ShowkaseBrowserApp`. Given that it uses Material 2, it shouldn't be too hard. There is [specific guidance](https://developer.android.com/about/versions/15/behavior-changes-15#not-edge-to-edge) on their docs on how to handle them correctly. 

The alternative is to migrate to Material 3 which handles insets out of the box with even less lines of code, but of course the migration to Material 3 requires more effort.

mataku:
## Context

Closes https://github.com/airbnb/Showkase/issues/399

TargetSDK 35 app requires edge-to-edge support. Of course, it's possible to opt out, but it's better to do it anyway because support is needed by TargetSDK 36, and contents will be natural if supported.

## Changes

- Support edge-to-edge in ShowkaseBrowserActivity and ShowkaseBrowserApp
  - Showkase uses material2, so I added safeDrawingPadding for composable component
- Set targetsdk to 35 in the sample project
- Apply enableEdgeToEdge to ensure that it applies to both OS 14 and below and 15

## Notes

I tried to change to use TopAppBar instead of `Row` use for toolbar, but I just found comments:

https://github.com/airbnb/Showkase/blob/cf3cdc80cec387442267866b2cde9400172c5034/showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt#L163-L167

TopAppBar seems to work, but ShowkaseBrowserTest failed, so I kept using custom Row.

## Screenshots

OS 14 | OS 14 with gesture nav 
:--: | :--: 
 |  | 

OS 15 | OS 15 with gesture nav 
:--: | :--:
 |


showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -124,46 +131,49 @@ internal fun ShowkaseAppBar(
...
+            )
+        }
     }
-
     /**
      * Commented out due to TopAppBar not working properly in beta-01 for this use case. Seems to be
      * related to use to Surface inside TopAppBar and a TextField. Creating my own implementation
-     * for now. Will uncomment if this issue gets fixed.
+     * for now. Wil
+     * l uncomment if this issue gets fixed.
```
olivierperez, line 176: I guess it is not wanted 🤔

mataku: 🤦 
Thanks! fixed: https://github.com/airbnb/Showkase/pull/412/commits/cba224c4082a4e3d00444fb3b3762db980c89952

olivierperez (commented):
I'm not a maintainer of this repository but I also would like this PR to be merged ^^

showkase/src/main/java/com/airbnb/android/showkase/ui/ShowkaseBrowserApp.kt
```diff
@@ -93,16 +92,20 @@ internal fun ShowkaseBrowserApp(
         LocalOnBackPressedDispatcherOwner provides backPressedDispatcherOwner
     ) {
         val navController = rememberNavController()
+        Surface(
+            color = Color.White
+        ) {
         Scaffold(
```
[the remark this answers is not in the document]
mataku: You're right. fixed: https://github.com/airbnb/Showkase/pull/412/commits/5890e6ed56c3557c80d9b5d9ee50cf2726c47364

vinaygaba:
@mataku Can you tell me what the test failure was when you attempted to use `TopAppBar`? This code path was written back in ~2021 before even the first stable release of Compose so the APIs were not as reliable as they are now. I'd love to use out of the box solutions wherever possible.

vinaygaba:
@mataku this rings a bell and iirc was an intentional choice to show longer component names. I think we can keep it as is and we should be good to merge once all the tests pass.

vinaygaba (approved):

mataku:
I agree. If it is changed, using the m3 AppBar would make a multi-line display much easier.

